PR #80, opened 2019-12-03
Replace aioboto3 with boto3

rahul003:
### Description of changes:
Removing aioboto3 dependency because it is becoming problematic to support it amidst all the conflicts with boto and botocore versions. Also addresses https://github.com/awslabs/sagemaker-debugger/issues/85 . Simply using boto3 with multiprocessing allows us to achieve even better performance than aioboto3 for our use case of downloading index files. No other S3 request is batched. Refreshing index files is one ListRequest, fetching a tensor is one GetObject

Todo:
- [x] Functionality testing
- [x] Perf testing
Screen Shot 2019-12-06 at 5 52 56 PM
- [x] Platforms checked are Mac, Linux. Had to incorporate a fix for Mac on py3.6. Read that windows with multiprocessing has problems in interactive environments, so disabled multiprocessing in windows.

#### Style and formatting:


setup.py
```diff
@@ -10,19 +10,7 @@
...
-    "aiohttp>=3.6.0,<4.0",  # aiobotocore breaks with 4.0
-    # boto3 explicitly depends on botocore
-    "boto3>=1.10.32",  # Sagemaker requires >= 1.9.213
-    "botocore>=1.13.32",
-    "nest_asyncio",
-    "protobuf>=3.6.0",
-    "numpy",
-    "packaging",
-]
+INSTALL_REQUIRES = ["protobuf>=3.6.0", "numpy", "packaging", "boto3>=1.10.32"]
```
Vikas-kum: can we pin all these versions to a unique version ?

smdebug/core/access_layer/s3handler.py
```diff
@@ -260,3 +267,176 @@ def close_client(self):
...
+            list_request = ListRequest(Bucket=delete_request.bucket, Prefix=delete_request.prefix)
+            keys = self.list_prefix(list_request)
+            if len(keys):
+                delete_dict = {}
+                delete_dict["Objects"] = []
+                for key in keys:
+                    delete_dict["Objects"].append({"Key": key})
+                self.client.delete_objects(Bucket=delete_request.bucket, Delete=delete_dict)
+
+    def close_client(self):
```
Vikas-kum: why the pass method? 
Should you close client here. and also create a destructor which closes the client.

NihalHarish:
Can you paste the results of the perf test here

smdebug/core/access_layer/s3handler.py
rahul003: I added it here to make the Handlers consistent for testing/benchmarking. Might remove it. For sync, there's nothing to do in close. Async, we have something to close.

rahul003:
I imagined for the restricted use case of downloading many index files at once, we might get away with only some performance regression. But I underestimated the advantage of async io even for this use case. Switching to boto3 doesn't really look like an option based on the following 10x performance boost we get with aioboto3.


| Bytes | Num files | Sync | Async |
| -- | -- | -- | -- |
| 10240 | 100 | 5.466304063796997 | 0.6316347122192383 |
|10240 | 1000 | 46.079941511154175 | 4.562414169311523 |
| 10240 | 10000 |457.6105201244354 | 59.778473138809204|

| Bytes | Num files | Sync | Async |
| -- | -- | -- | -- |
| 10240 | 1 | 0.0649409294128418 | 0.09938454627990723 |
| 10240 | 10 | 0.2014918327331543 | 1.1134459972381592 |
| 10240 | 100 | 2.155343532562256 | 0.3696606159210205 |
|10240 | 1000 | 24.51427435874939 | 3.7332723140716553 |


| Bytes | Num files | Sync | Sync with s3 transfer | Async |
| -- | -- | -- | -- | -- |
| 10240 | 1 | 0.17870569229125977 | 0.07202434539794922 | 0.10649681091308594 |
| 10240 | 10 | 0.1648552417755127 | 0.29348111152648926 | 0.07737946510314941 |
| 10240 | 100 | 2.131474018096924 | 3.889939308166504 | 1.1508255004882812 |
|10240 | 1000 | 25.46154403686523 | 49.814311027526855 | 4.341756343841553 |

jarednielsen:
Oof, looks like aioboto3 provides nearly linear scaling (assuming you're running on a 4-core, 8-thread machine). I wonder if we can achieve similar gains natively by using Python multiprocessing with boto3.

rahul003:
I was running on a p3.8x but it shouldn’t matter as I’d have been using only 1 core anyway. I’d be curious to see how multi processing compares. But make sure to include the time to send data back to the main process in the timing, as this data transfer can be expensive. Using shared memory instead of copying from process to process would help

Vikas-kum:
> Oof, looks like aioboto3 provides nearly linear scaling (assuming you're running on a 4-core, 8-thread machine). I wonder if we can achieve similar gains natively by using Python multiprocessing with boto3.
> 
> Getting some initial promising results (at home with a slower network):
> Bytes 	Num files 	Sync with multiprocessing (32 threads) 	Async
> 10240 	1 	0.87 	0.67
> 10240 	10 	1.11 	0.75
> 10240 	100 	1.74 	1.83
> 10240 	1000 	5.65 	12.28
> 10240 	3000 	16.59 	45.49
> 10240 	10000 	46.43 	148.38
> 
> This approach seems significantly faster, actually. I'll do some further testing tomorrow on a faster network to see if it holds up.
> 
> Results from a faster network:
> Bytes 	Num files 	Sync with multiprocessing (64 threads) 	Async
> 10240 	1 	0.71 	0.49
> 10240 	10 	0.96 	0.34
> 10240 	100 	2.14 	0.91
> 10240 	1000 	4.30 	10.11
> 10240 	3000 	7.54 	30.17
> 10240 	10000 	21.28 	109.12

Can you check why Rahul's number is 4.5 s for 1000 files and yours is 10.11 .
Also, try import numpy in the test code. These library probably set cpu affinity, so we should check that nothing weird happens with multiprocessing in those cases.
You can see the speedup with multiprocessing, which probably comes at cost of cpu utilization.

rahul003:
I think Jared is running tests on his Mac. So the Mac vs EC2 CPU difference I'm assuming is the reason behind the slower aiobot3 performance in his case

rahul003:
Did some more benchmarking, here are the results. Will add description in some time. https://quip-amazon.com/TZggAhnoJCn4

setup.py
rahul003: Python dependency management works different than other languages, due to how PyPI dependency resolution works. 

As library developers, we should not be pinning to any specific version. The maintainers of environments which include this library, or users of library can pin to specific versions of dependencies they use. If the library itself pins a specific version of a dependency, then some other package which also depends on that same dependency in the same environment can't be installed due to version conflicts. 

Here's a reference https://stackoverflow.com/a/44938662

rahul003:
CI for this PR will not pass as we need to make changes in sagemaker-debugger-rules as well. The CI has passed all tests for this repository, but failing on rules. PR has to be merged in first, and then the rules repository needs to be updated to remove its usage of aioboto3 and update its usage of S3Handler.

setup.py
Vikas-kum: We should not block this PR.  Let's talk about pinning, goals behind it and how can we achieve those goals.

PR #91, opened 2019-12-04
Rename env variable

NihalHarish:
### Description of changes:
- rename variable from  TORNASOLE_EVENT_FILE_RETRY_LIMIT to MISSING_EVENT_FILE_RETRY_LIMIT
- Add docs to api.md
#### Style and formatting:


smdebug/core/config_constants.py
```diff
@@ -31,7 +31,7 @@
 
 INCOMPLETE_STEP_WAIT_WINDOW_KEY = "INCOMPLETE_STEP_WAIT_WINDOW"
 INCOMPLETE_STEP_WAIT_WINDOW_DEFAULT = 1000
-DEFAULT_EVENT_FILE_RETRY_LIMIT = 100
```
rahul003: Define the key here

docs/api.md
```diff
@@ -349,3 +349,11 @@ directory contains collections, events, and index files. A trial checks to see i
 specified in the smdebug hook has been completed. This environment variable
 specifies the maximum number of incomplete steps that the trial will wait for before marking
 half of them as complete. Default: 1000
+
```
rahul003: I feel like all these env variables should have a SMD_ prefix

rahul003: This is technically a backwards incompatible change, but I think we can make an exception here as these were `undocumented` before. But just to be sure, can you add a note at the top saying these are applicable for versions > 0.4.14

docs/api.md
```diff
@@ -437,6 +437,68 @@ abs_l1
 abs_l2
 ```
 
+#### `TENSORBOARD_CONFIG_FILE_PATH`:
```
rahul003: Wait these should be moved to env_var.md file. I guess the merge conflict wasn't resolved correctly

PR #97, opened 2019-12-06
cleaned up examples folder a bit

ddavydenko:
### Description of changes:
- removed last notebook from /examples/mxnet
- synced readme files in /examples folders
- put up one script for mxnet demo of creating hook in script

#### Style and formatting:


examples/xgboost/README.md
```diff
@@ -1,2 +0,0 @@
-## Example Notebooks
```
rahul003: Was removing this intentional?

ddavydenko:
There is no docs for XGBoost here currently, so I thought no point in keeping just a read me with external link. 

—
Thanks,
Denis

examples/mxnet/README.md
```diff
@@ -1,2 +1,22 @@
-## Example Notebooks
+# Examples
+## Example notebooks
 Please refer to the example notebooks in [Amazon SageMaker Examples repository](https://github.com/awslabs/amazon-sagemaker-examples/tree/master/sagemaker-debugger)
+
+## Example scripts
```
ddavydenko:
```suggestion
## Example scripts 
```

examples/mxnet/README.md
```diff
@@ -1,2 +1,22 @@
...
```
ddavydenko:
```suggestion
## Example scripts
```

examples/xgboost/README.md
ddavydenko: There is no docs for XGBoost here currently, so I thought no point in keeping just a read me with external link.

rahul003: We do have a doc page for XGBoost though, I think it would be nice if we have this page just to be clearer.

examples/mxnet/scripts/mxnet_fashionmnist_w_g_b.py
```diff
@@ -1,7 +1,14 @@
+"""
+This script is a simple FashionMNIST training script which uses MXNet's.
+It has been orchestrated with SageMaker Debugger hook to allow saving tensors during training.
+Here, the hook has been created using its constructor to allow running this locally for your experimentation.
+When you want to run this script in SageMaker, it is recommended to create the hook from json file.
+Please see scripts in either /examples/tensorflow/sagemaker_byoc or /examples/tensorflow/sagemaker_official_container
```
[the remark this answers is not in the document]
ddavydenko: Why would not we want to point to examples that demonstrate other use cases, even though they are not using same engine?

ddavydenko:
> LGTM.
> Some of these examples are used for testing 1P rules. I noticed these examples are already moved to 1P rules repository? The 1P Rule CI need to pick the examples from that repo after this change.

CI already passes on setup of MXNet tests introduced in this PR so I don't think rules CI will be a problem.

examples/mxnet/scripts/mxnet_fashionmnist_w_g_b.py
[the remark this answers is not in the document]
rahul003: Okay since we don't have such an example for MXNet, I see your point

rahul003:
Rules CI has the previous scripts copied into their repo

PR #115, opened 2019-12-16
Pinning the version of numpy that works with the supported versions o…

leleamol:
…f DL Frameworks

### Description of changes:
The DL Framework versions that smdebug currently supports require numpy version between 1.16 and 2.0). We can pin the numpy version to '>=1.16 and <2.0' 

In future, if the dependency of DL Frameworks on numpy becomes conflicting, we can consider creating separate wheels for DL frameworks.

#### Style and formatting:


Vikas-kum:
There is one problem with pinning the upper version. 
https://github.com/boto/botocore/issues/1872 - Look at this, where they pinned python-dateutil to 2.8.1 and there are so many people unhappy as it started breaking things.

jarednielsen:
69 unhappy customers!

leleamol:
@Vikas-kum @jarednielsen  Tensorflow and mxnet are also following the same strategy.

https://github.com/tensorflow/tensorflow/blob/d74412af53590e18a07d8cc5cfb042efbf42fd73/tensorflow/tools/pip_package/setup.py#L61

smdbeug will be used along with one of these frameworks.   Ideally, we should limit the pinning of dependencies  to the framework versions but that's not possible. Hence we should use the packages that will be used by framework versions that we currently support.

leleamol:
@Vikas-kum 
I am not getting why pinning the numpy version is not a good idea. If it is acceptable for the TF and MXnet versions that we are supporting, it should be ok for us too. 
 Are there any unhappy customers of TF due to pinning of numpy?

Vikas-kum:
> @Vikas-kum
> I am not getting why pinning the numpy version is not a good idea. If it is acceptable for the TF and MXnet versions that we are supporting, it should be ok for us too.
> Are there any unhappy customers of TF due to pinning of numpy?

Look at the botocore pinning issue where they pinned upper version. We dont want people to get in such a mess  which can be created by pinning upper versions.

leleamol:
> > @Vikas-kum
> > I am not getting why pinning the numpy version is not a good idea. If it is acceptable for the TF and MXnet versions that we are supporting, it should be ok for us too.
> > Are there any unhappy customers of TF due to pinning of numpy?
> 
> Look at the botocore pinning issue where they pinned upper version. We dont want people to get in such a mess which can be created by pinning upper versions.

@Vikas-kum Are we certain that smdebug works or will work if numpy is upgraded to 2.0?
We can always remove this upper version bound later.  But IMO we should be conservative and pin the upper and lower bounds both. 
With the framework versions that we are currently supporting, we are ok with pinning upper and lower versions of numpy. If any of these frameworks start supporting numpy 2.0, they will do so with the new updated version of that framework.  In that case, we would have to evaluate whether smdebug supports that version of not. At that time, we can reconsider removing the upper bound on numpy.
@ddavydenko what are your thoughts?

PR #116, opened 2019-12-20
Nightly Build setup

anirudhacharya:
### Description of changes:

changes to enable nightly build.

#### Style and formatting:


setup.py
```diff
@@ -1,13 +1,14 @@
...
 import os
 import sys
+from datetime import date
 
 # Third Party
 import setuptools
 
-exec(open("smdebug/_version.py").read())
+# First Party
+import smdebug
```
jarednielsen: We can't import smdebug in setup.py - this is where we build the package in the first place. Why are we switching this around?

anirudhacharya: i can switch it back, but this works too. i have checked.

setup.py
```diff
@@ -29,8 +46,10 @@ def build_package(version):
     setuptools.setup(
         name="smdebug",
         version=version,
+        long_description="\n".join(DOCLINES[2:]),
+        long_description_content_type="text/x-rst",
         author="AWS DeepLearning Team",
-        description="Automated debugging for machine learning",
+        description=DOCLINES[0],
```
jarednielsen: What are we changing this to? Someone might change the README later and unintentionally mess up the description. Might be best to just leave the description hard-coded.

setup.py
jarednielsen: Interesting. As long as you've tested it in a fresh environment where smdebug isn't already installed - just never seen it done this way before.

setup.py
```diff
@@ -82,4 +101,15 @@ def git(*args):
 
     sys.exit(1)
 
-build_package(version=CURRENT_VERSION)
+
+def detect_smdebug_version():
+    sys.path.append(os.path.abspath("smdebug"))
```
jarednielsen: Why is this necessary?

setup.py
anirudhacharya: this change was done on Amol's request. @leleamol

setup.py
anirudhacharya: removed.

setup.py
```diff
@@ -1,13 +1,30 @@
+#!/usr/bin/env python
+""" Amazon SageMaker Debugger is an offering from AWS which help you automate the debugging of machine learning training jobs.
```
Vikas-kum: s/help/helps

PR #117, opened 2019-12-25
Fix for MXNet integ test

ddavydenko:
### Description of changes:
This PR fixes ZCC integ test for MXNet as it is not runnable as is. It uses SagemakerSimulator to ensure that ZCC file is created for the test and also changes save_interval to 100 as with 0.5 version default save_interval is 500, so training job in this test runs only for 1 epoch which generates less than 300 steps.

#### Style and formatting:

#### Testing performed:
I pulled in official MXNet training DLC that has ZCC MXNet version in it (763104351884.dkr.ecr.us-east-1.amazonaws.com/mxnet-training:1.6.0-cpu-py36-ubuntu16.04-v3.1), volume mounted it with access to sagemaker-debugger code repo and ran this to verify MXNet integ test is passing:
```shell
cd /sagemaker-debugger
python tests/zero_code_change/mxnet_gluon_integration_test.py
```

### DIFF
--- a/tests/zero_code_change/mxnet_gluon_integration_test.py
+++ b/tests/zero_code_change/mxnet_gluon_integration_test.py
@@ -8,6 +8,9 @@ import numpy as np
 from mxnet import autograd, gluon
 from mxnet.gluon import nn
 
+# First Party
+from smdebug.core.utils import SagemakerSimulator
+
 
 def parse_args():
     parser = argparse.ArgumentParser(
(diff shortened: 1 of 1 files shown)


ddavydenko:
Sorry, I missed providing instructions on what was tested. Please find them in description.

I do realize that CI is currently failing for some tests, however this PR is meant to address only one failure - for MXNet integ test. I wonder if it can be merged even without other tests passing.

PR #125, opened 2020-01-02
0.5

ddavydenko:
### Description of changes:
Pulling in changes from 0.5 into master (project description for pypi publishing is the only new thing, everything else is already in master)


smdebug/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.5.0"
+__version__ = "0.5.0.post0"
```
anirudhacharya: nit: new line

why this change? in the nightly pre-build release, we fetch this version number append 'b' 'timestamp' to generate the pre-build version. this upload upload might now fail since `0.5.0.post0b<timestamp>` might not meet semver standards of pypi.

setup.py
```diff
@@ -25,6 +25,7 @@
 import smdebug
 
 DOCLINES = (__doc__ or "").split("\n")
+CURRENT_VERSION = __version__
```
anirudhacharya: are we using this anywhere?

smdebug/_version.py
ddavydenko: this was because 0.5.0 publish happened without description for pypi so next version after that is 0.5.0.post0.

setup.py
ddavydenko: not sure... I was under impression that is used but if it is not - happy to remove it.

PR #137, opened 2020-01-08
Fixes: SMSimulator fix, listing files local should ignore tmp files

Vikas-kum:
### Description of changes:
1. SMSimulator needed fix for running tf zcc test end to end
2. Fix for list directory to ignore tmp files.

#### Style and formatting:

### DIFF
--- a/smdebug/core/index_reader.py
+++ b/smdebug/core/index_reader.py
@@ -351,12 +351,12 @@ class LocalIndexReader(IndexReader):
 
     def list_index_files(self):
         index_dirname = IndexFileLocationUtils.get_index_path(self.path)
-        index_files_regex = ".+\/\d+_(.+)\.(json|csv)$"
+        index_files_regex = "\d+_(.+)\.(json|csv)$"
         index_files = list_files_in_directory(index_dirname, file_regex=index_files_regex)
         return sorted(index_files)
 
     def list_event_files(self, start_after_key=None):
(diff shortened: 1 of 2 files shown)


smdebug/core/utils.py
```diff
@@ -219,6 +221,11 @@ def get_tb_worker():
     return f"{os.getpid()}_{socket.gethostname()}"
 
 
+def remove_file_if_exist(file_path):
```
jarednielsen: Nit: use `exists` to be consistent with other methods

smdebug/core/index_reader.py
```diff
@@ -351,11 +351,15 @@ def _is_event_file_present(self, file):
 
     def list_index_files(self):
         index_dirname = IndexFileLocationUtils.get_index_path(self.path)
-        index_files = list_files_in_directory(index_dirname)
+        index_files_regex = ".+\/\d+_(.+)\.(json|csv)$"
```
jarednielsen: Add a comment explaining the regex. I understand filtering so it only gets `.json` or `.csv`, but it's not as clear filtering capturing groups with an underscore in the middle. Just worried that someone will change the file naming format and we'll have a bunch of silently ignored files.

Vikas-kum: Put comment. Made regex more relaxed so look for .json, .csv or .tfevents

PR #142, opened 2020-01-10
Continuing rule execution in case we get TensorUnavailable exception

Vikas-kum:
As rule might be looking for tensor which may be saved in future step,  in case we get TensorUnavailable , invoker will go on with executing rules.

#### Style and formatting:

### DIFF
--- a/smdebug/rules/rule_invoker.py
+++ b/smdebug/rules/rule_invoker.py
@@ -1,6 +1,11 @@
 # First Party
 from smdebug.core.logger import get_logger
-from smdebug.exceptions import RuleEvaluationConditionMet, StepUnavailable, TensorUnavailableForStep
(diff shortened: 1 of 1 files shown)


jarednielsen:
What's the difference between `TensorUnavailable` and `TensorUnavailableForStep`? Might be good opportunity for consolidation so we don't miss catching them like we had here.

Vikas-kum:
> `TensorUnavailable`

https://github.com/awslabs/sagemaker-debugger/blob/master/docs/analysis.md#exceptions

TensorUnavailableForStep , means tensor was returned in tensor_names() , so it is present for some step , but not the one you asked in tensor.value(step=5)

TensorUnavailable --> Tensor is still to be seen.

jarednielsen:
Got it. Lgtm

PR #158, opened 2020-01-24
Path String Sanitation

NihalHarish:
### Description of changes:
- Remove trailing, leading whitespaces in path string input to create_trial
- Make warning log emitted in `smdebug/trials/trial.py` print out the input path str, to help debugging
- Moved two helper functions: `dummy_step_creator` and `dummy_trial_creator` to utils to allow reuse

#### Style and formatting:


rahul003:
Why not just error out? 
Training and leading are still fine, but interspersed seems like we're trying too hard to fix user's mistakes. What if by removing that space in between we're concatenating a string which the user didn't intend to pass as path?

NihalHarish:
> Why not just error out?
> Training and leading are still fine, but interspersed seems like we're trying too hard to fix user's mistakes. What if by removing that space in between we're concatenating a string which the user didn't intend to pass as path?

Hmm, you're right, interspersed isn't required.

PR #162, opened 2020-02-05
Raise Error For Invalid Collection Config

NihalHarish:
### Description of changes:
- Raise InvalidCollectionConfiguration exception if the customer has not correctly configured a collection
- Motivation behind the change, customers may make spelling mistakes or forget to correctly configure their custom collections.
- We want to enforce correct configurations so that they do not have to bear the cost of an entire job running with incorrect collection configurations
* Changes:
  * Added a new Exception class `InvalidCollectionConfiguration` in `smdebug/exceptions.py`
  * Added a check in `_prepare_collections` in `smdebug/core/hook.py`
  * Added a test `test_invalid_collection_config_exception`  in `tests/core/test_collections.py`

#### Style and formatting:

### DIFF
--- a/smdebug/core/collection.py
+++ b/smdebug/core/collection.py
@@ -46,8 +46,33 @@ class CollectionKeys:
     TREES = "trees"
 
     @classmethod
-    def values(cls):
-        return [value for name, value in vars(cls).items() if name.isupper()]
+    def builtins(cls, framework=None):
+        default = [
+            CollectionKeys.DEFAULT,
+            CollectionKeys.ALL,
+            CollectionKeys.INPUTS,
+            CollectionKeys.OUTPUTS,
+            CollectionKeys.WEIGHTS,
+            CollectionKeys.GRADIENTS,
+            CollectionKeys.LOSSES,
+            CollectionKeys.BIASES,
+            CollectionKeys.SM_METRICS,
+            CollectionKeys.OPTIMIZER_VARIABLES,
+            CollectionKeys.TENSORFLOW_SUMMARIES,
+        ]
+        if framework is None:
+            return default
+        elif framework == "xgboost":
+            return [
+                CollectionKeys.HYPERPARAMETERS,
+                CollectionKeys.METRICS,
+                CollectionKeys.PREDICTIONS,
+                CollectionKeys.LABELS,
+                CollectionKeys.FEATURE_IMPORTANCE,
+                CollectionKeys.AVERAGE_SHAP,
+                CollectionKeys.FULL_SHAP,
+                CollectionKeys.TREES,
+            ] + default
 
 
 # Collection with summary objects instead of tensors
--- a/smdebug/core/collection.py
+++ b/smdebug/core/collection.py
@@ -45,6 +45,10 @@ class CollectionKeys:
     FULL_SHAP = "full_shap"
     TREES = "trees"
 
+    @classmethod
+    def values(cls):
+        return [value for name, value in vars(cls).items() if name.isupper()]
+
 
 # Collection with summary objects instead of tensors
 # so we don't create summaries or reductions of these
--- a/smdebug/core/collection.py
+++ b/smdebug/core/collection.py
@@ -34,45 +34,16 @@ class CollectionKeys:
 
     OPTIMIZER_VARIABLES = "optimizer_variables"
     TENSORFLOW_SUMMARIES = "tensorflow_summaries"
+    METRICS = "metrics"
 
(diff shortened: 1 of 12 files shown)


smdebug/core/collection.py
```diff
@@ -45,6 +45,10 @@ class CollectionKeys:
     FULL_SHAP = "full_shap"
     TREES = "trees"
 
+    @classmethod
+    def values(cls):
+        return [value for name, value in vars(cls).items() if name.isupper()]
```
rahul003: Risky to do it this way. Predictions for example is a valid collection user might want to create for TF/PT/MX. But since it's in the 1p collections for xgboost, your check won't be applied for that.

rahul003: Also, can you add a test for this values method. What else can vars() return

tests/core/test_collections.py
```diff
@@ -87,6 +88,7 @@ def test_collection_defaults_to_hook_config():
   """
     cm = CollectionManager()
     cm.create_collection("foo")
+    cm.get("foo").include_regex = "*"
```
vandanavk: can the string be empty too? regex='*' would include ''

tests/core/test_collections.py
```diff
@@ -101,3 +103,28 @@ def test_collection_defaults_to_hook_config():
...
+    )
+    hook.collection_manager = cm
+    try:
+        hook._prepare_collections()
+    except InvalidCollectionConfiguration:
+        pass
+    else:
+        assert False, "Invalid Collection Name did not raise error"
+
+    cm.get("foo").include_regex = "*"
```
vandanavk: allows empty string here as well

smdebug/pytorch/hook.py
```diff
@@ -103,15 +103,18 @@ def _log_params(self, module):
...
     def _prepare_collections(self):
-        super()._prepare_collections()
         for coll in self.collection_manager.collections.values():
             for m, (include_inputs, include_outputs) in coll.modules.items():
                 module_name = self.module_maps[m]
                 if include_inputs:
                     coll.include(module_name + "_input_")
                 if include_outputs:
                     coll.include(module_name + "_output_")
+        super()._prepare_collections()
```
vandanavk: why this move?

smdebug/mxnet/collection.py
```diff
@@ -21,6 +22,8 @@ def __init__(self, create_default=True):
             self._register_default_collections()
 
     def _register_default_collections(self):
+        for c in DEFAULT_MXNET_COLLECTIONS:
+            self.create_collection(c)
```
vandanavk: do we need these loops in all _register_default_collections()? collectionmanager.get() creates the collection (if not available) by default. Collection.ALL and DEFAULT are created in core/hook.py and the framework specific _register_default_collections() create their respective default collections.

tests/core/test_collections.py
```diff
@@ -101,3 +103,28 @@ def test_collection_defaults_to_hook_config():
     hook._prepare_collections()
     assert cm.get("foo").save_config.mode_save_configs[ModeKeys.TRAIN].save_interval == 10
     assert cm.get("foo").reduction_config.save_raw_tensor is True
+
+
+def test_invalid_collection_config_exception():
```
vandanavk: are there any failures in the existing tests? like the rules tests and scalar tests?

smdebug/core/collection.py
NihalHarish: Removed.

tests/core/test_collections.py
NihalHarish: The user cannot input an empty string, "*" however is accepted.

tests/core/test_collections.py
NihalHarish: "*" is valid

smdebug/pytorch/hook.py
NihalHarish: `super()._prepare_collections()` checks if include_regex has been set, but this happens in lines 108-116, so we need to move the fn call

tests/core/test_collections.py
NihalHarish: There were failures in the rules repository, which were addressed in a  separate [PR](https://github.com/awslabs/sagemaker-debugger-rules/pull/194)

smdebug/mxnet/collection.py
NihalHarish: This is redundant but we just want to enforce this style where we separate the creation step from the get call

PR #165, opened 2020-02-15
Pytorch data parallel bug

NihalHarish:
### Description of changes:
- Bug has been documented [here](https://answers.amazon.com/questions/93236)
- Removed the module_map datastructure and replaced it with a module_set which prevents the registration of modules more than once
- Assigned module names to the module objects themselves so that the attribute value is available across devices.
- This is a partial fix that allows customers to use the Pytorch Dataparallel construct with encountering bugs, next steps would be the creation a separate collection/index/event files for each worker

#### Style and formatting:

### DIFF
--- a/tests/pytorch/test_data_parallel.py
+++ b/tests/pytorch/test_data_parallel.py
@@ -40,9 +40,8 @@ def get_dataloader():
 def test_data_parallel():
     device = "cuda" if torch.cuda.is_available() else "cpu"
     model = Net()
-    if device == "cpu":
-        model = model.to(device)
-    else:
+    model = model.to(device)
+    if device == "cuda":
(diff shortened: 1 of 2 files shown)


tests/pytorch/test_data_parallel.py
```diff
@@ -0,0 +1,43 @@
...
+import torch
+import torch.optim as optim
+from tests.pytorch.utils import Net, train
+from torch.nn.parallel import DataParallel
+
+# First Party
+import smdebug.pytorch as smd
+from smdebug.trials import create_trial
+
+out_dir = "/tmp/run"
```
larroy: Can we use TempDir instead of hardcoded directories? This is not portable.

Issue #161, opened 2020-01-28
create_trial breaks


Vikas-kum:
>>> t =create_trial('./output')
[2020-01-27 19:31:32.274 88e9fe53272d.ant.amazon.com:84283 INFO local_trial.py:35] Loading trial output at path ./output
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/vikumar/anaconda3/lib/python3.6/site-packages/smdebug-0.6.0b20200127-py3.6.egg/smdebug/trials/utils.py", line 19, in create_trial
  File "/Users/vikumar/anaconda3/lib/python3.6/site-packages/smdebug-0.6.0b20200127-py3.6.egg/smdebug/trials/local_trial.py", line 37, in __init__
  File "/Users/vikumar/anaconda3/lib/python3.6/site-packages/smdebug-0.6.0b20200127-py3.6.egg/smdebug/trials/trial.py", line 556, in _load_tensors
  File "/Users/vikumar/anaconda3/lib/python3.6/site-packages/smdebug-0.6.0b20200127-py3.6.egg/smdebug/trials/trial.py", line 621, in _load_tensors_from_index_files
  File "/Users/vikumar/anaconda3/lib/python3.6/site-packages/smdebug-0.6.0b20200127-py3.6.egg/smdebug/core/index_reader.py", line 403, in load_tensor_data_from_index_files
  File "/Users/vikumar/anaconda3/lib/python3.6/site-packages/smdebug-0.6.0b20200127-py3.6.egg/smdebug/core/index_reader.py", line 448, in read_index_files
  File "/Users/vikumar/anaconda3/lib/python3.6/site-packages/smdebug-0.6.0b20200127-py3.6.egg/smdebug/core/index_reader.py", line 85, in add
  File "/Users/vikumar/anaconda3/lib/python3.6/site-packages/smdebug-0.6.0b20200127-py3.6.egg/smdebug/core/index_reader.py", line 79, in _evict_cache
TypeError: '<' not supported between instances of 'str' and 'NoneType'

PR #168, opened 2020-02-26
Fix create_trial brk when many elements are read in the first attempt

NihalHarish:
### Description of changes:
- If more than self.cache_limit number of files are read in the first attempt, an incorrect check between start_after_key and elements present in the cache caused an exception.
- This PR sets a default string value when start_after_key is `None`

#### Style and formatting:

#161 

### DIFF
--- a/tests/core/test_utils.py
+++ b/tests/core/test_utils.py
@@ -98,6 +98,17 @@ def test_index_files_cache():
     )  # Elements in the cache will be file_4, file_5, file_6
 
 
+def test_index_files_cache_insert_many_elements_in_the_first_read():
+    cache = ReadIndexFilesCache()
+    cache.cache_limit = 5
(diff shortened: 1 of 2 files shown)


tests/core/test_utils.py
```diff
@@ -98,6 +98,17 @@ def test_index_files_cache():
...
 
 
+def test_index_files_cache_insert_many_elements_in_the_first_read():
+    cache = ReadIndexFilesCache()
+    cache.cache_limit = 5
+    elements = ["a", "b", "c", "d", "e", "f", "g", "h"]
+    for e in elements:
+        cache.add(e, None)
+
+    # No files should be evicted because start_after_key has not been set
```
larroy: this is kind of weird. Is it by design like this?

PR #170, opened 2020-02-27
Update the save_scalar() method to accept and store the timestamp at along with the scalar value

leleamol:
### Description of changes:
The current implementation of save_scalar() only caches the scalar values. The scalar values actually get saved/stored on a disk when the next step is written.
The wall_time corresponding to the tfevent represents time at which data is getting stored on the disk. It does not represent the time when user has generated the scalar data.

The current change supports following:

1. It allows user to specify the timestamp along with the scalar value. The given timestamp will be cached and then later saved to disk when the next step is written.
2. If user does not specify the timestamp with save_scalar(), the timestamp at the time when this method is called will be cached and then later saved.

The testing is done as follows:

1. The testscript stores various scalars using save_scalar() method. The script records the time stamp and value of scalar that is saved.
2. After the training, the scalars are read from tfevent files.
3. The test ensures that all the scalars are available in the tfevent files. The timestamps read from tfevents match the timestamp that were recorded while training. The values of scalars read from tfevents match the values of scalars recorded during training.

#### Style and formatting:


smdebug/core/reader.py
```diff
@@ -54,3 +57,23 @@ def read_tensors(self, check="minimal"):
         if check is True:
             check = "minimal"
         return self._reader.read_tensors(check=check)
+
+    # Read the events that match the given regex list. If Regex list is not specified all events are read.
+    def read_events(self, check="minimal", regex_list=None):
+        if check is True:
+            check = "minimal"
```
anirudhacharya: is it okay to have such weakly typed semantics, where we are changing the type of the variable from boolean to string?

larroy: Agree with @anirudhacharya We should avoid this kind of pattern.

leleamol: @anirudhacharya @larroy 
Thanks for catching it!
Will fix it.

leleamol: Addressed the review comments.

smdebug/core/reader.py
```diff
@@ -54,3 +57,23 @@ def read_tensors(self, check="minimal"):
         if check is True:
             check = "minimal"
         return self._reader.read_tensors(check=check)
+
+    # Read the events that match the given regex list. If Regex list is not specified all events are read.
+    def read_events(self, check="minimal", regex_list=None):
```
larroy: please add docstring instead of comment on top, so the function is properly documented. Is not clear what the check "minimal argument is doing. I know the code was there before, but since you are now modifying it is a good opportunity to improve it if only slightly.

tests/core/utils.py
```diff
@@ -1,11 +1,167 @@
...
 
 
 def write_dummy_collection_file(trial):
     cm = CollectionManager()
     cm.create_collection("default")
     cm.add(Collection(trial))
     cm.export(trial, DEFAULT_COLLECTIONS_FILE_NAME)
+
+
+def delete_local_trials(local_trials):
```
larroy: docstring missing.

This looks kind of dangerous. Is there a way that we can check that they are in fact, trials and not just '/'

tests/core/utils.py
```diff
@@ -1,11 +1,167 @@
...
+    for trial in local_trials:
+        shutil.rmtree(trial)
+
+
+# need this to seek to the right file offset for test output verification
+metrics_file_position = 0
```
larroy: why is this global? Can't we make it a class?

tests/core/utils.py
```diff
@@ -1,11 +1,167 @@
...
+            assert len(set(save_steps["TRAIN"]) & set(train_metric[train_data])) == len(
+                save_steps["TRAIN"]
+            )
+        for eval_data in eval_metric:
+            assert len(set(save_steps["EVAL"]) & set(eval_metric[eval_data])) == len(
+                save_steps["EVAL"]
+            )
+
+
+def check_trials(out_dir, save_steps, saved_scalars=None):
```
larroy: can we improve the naming? check doesn't mean much, I would suggest "ensure written" or similar.

tests/core/utils.py
```diff
@@ -1,11 +1,167 @@
...
+def delete_local_trials(local_trials):
+    for trial in local_trials:
+        shutil.rmtree(trial)
+
+
+# need this to seek to the right file offset for test output verification
+metrics_file_position = 0
+
+
+def check_metrics_file(save_steps, saved_scalars=None):
```
larroy: better naming?

tests/core/utils.py
```diff
@@ -1,11 +1,167 @@
...
+            return
+        file_name = "{}/{}.json".format(METRICS_DIR, str(os.getpid()))
+        scalarnames = set()
+
+        import collections
+
+        train_metric = collections.defaultdict(list)
+        eval_metric = collections.defaultdict(list)
+
+        with open(file_name) as fp:
```
larroy: can we break into smaller functions?

tests/core/utils.py
larroy: If it's a private module function, prepend with underscore as per python convention,

tests/mxnet/test_hook_save_scalar_mxnet.py
```diff
@@ -0,0 +1,129 @@
...
+
+# First Party
+from smdebug.core.modes import ModeKeys
+from smdebug.core.save_config import SaveConfig, SaveConfigMode
+from smdebug.mxnet import Hook as MX_Hook
+
+SMDEBUG_MX_HOOK_TESTS_DIR = "/tmp/test_output/smdebug_mx/tests/"
+
+
+def simple_mx_model(hook, steps=10, register_loss=False, with_timestamp=False):
```
larroy: steps -> epochs?
also below.

smdebug/core/hook.py
```diff
@@ -48,13 +49,16 @@
 
 
 class ScalarCache(object):
```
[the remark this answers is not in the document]
leleamol: Added

smdebug/core/reader.py
leleamol: Added

tests/core/utils.py
leleamol: Added the docstring. However, these functions are utility functions for testing. They are not API. As per the current code base, it is getting used in current test case only.

tests/core/utils.py
leleamol: Fixed.

tests/mxnet/test_hook_save_scalar_mxnet.py
leleamol: The steps is correct in this context.  For this test, we are training the model for few steps only not the entire epoch.

tests/core/utils.py
leleamol: We can address the refactoring and renaming of functions in different PR for code cleanup.
These particular functions are utility functions in test code and currently only getting used in save_scalar related test only.
If we encounter more usage for these functions, we will definitly reorganize them.

smdebug/core/hook.py
```diff
@@ -48,13 +49,29 @@
...
-    def __init__(self, scalar_name, scalar_val, mode, sm_metric, write_tb, write_event):
+    def __init__(
```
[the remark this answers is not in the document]
leleamol: @vandanavk might be able to answer this question correctly.  The ScalarCache object is not part of public facing API.  It is a cache created by save_scalar() method. The external users are not expected to create this object.  IMO this could be the reason.

smdebug/core/hook.py
```diff
@@ -621,7 +642,7 @@ def _write_raw_tensor(self, tensor_name, tensor_value, save_collections, tensor_
                 self._write_raw_tensor_simple(tensor_name, tensor_value, tensor_ref=tensor_ref)
                 break
 
-    def _write_raw_tensor_simple(self, tensor_name, tensor_value, tensor_ref=None):
+    def _write_raw_tensor_simple(self, tensor_name, tensor_value, tensor_ref=None, timestamp=None):
```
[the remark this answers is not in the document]
leleamol: @anirudhacharya write_raw_tensor() writes the scalar tensors to events file in tfevents format.
For current use case, I am updating the methods that are writing scalars only.

If needed we can implement it for generic tensors later

PR #174, opened 2020-03-09
Update the comments for 'out_dir' for hook class.

leleamol:
### Description of changes:
The hook does not support merging the tensors from current job with tensors from previous jobs.
Therefore, hook initializing routine throws an error if the directory specified by 'out_dir' exists.

Since this behavior is not documented in readme, users may try to instantiate the hook with the same 'out_dir' path. 

#### Style and formatting:


larroy:
please rebase.

PR #178, opened 2020-03-13
Use executing_eagerly_outside_functions for global execution context

vandanavk:
### Description of changes:
TF 2.x keras switches to non-eager mode during forward pass and tf.executing_eagerly returns the current context executing_eagerly=False. Relying on this check in the case of ZCC, results in an error when include_collections = [gradients].

All other tf.executing_eagerly() calls are maintained as it is, as the callback hooks are called outside tf.function scope.

Testing done:
Add more tests in ZCC integration test to catch this case.
The issue doesn't occur with vanilla TF because set_gradients() is not called - added tests for vanilla TF anyway

#### Style and formatting:


tests/zero_code_change/tensorflow2_integration_tests.py
```diff
@@ -77,6 +80,103 @@ def test_keras_v2(script_mode: bool = False, eager_mode: bool = True):
         trial = smd.create_trial(path=sim.out_dir)
         assert len(trial.steps()) > 0, "Nothing saved at any step."
         assert len(trial.tensor_names()) > 0, "Tensors were not saved."
+        assert len(trial.tensor_names(collection="losses")) > 0
+
+
+def helper_test_keras_v2_json_config(
+    json_file_contents, script_mode: bool = False, eager_mode: bool = True
+):
+    """ Works as intended. """
```
jarednielsen: Can you improve this comment to describe what is intended?

tests/zero_code_change/tensorflow2_integration_tests.py
```diff
@@ -77,6 +80,103 @@ def test_keras_v2(script_mode: bool = False, eager_mode: bool = True):
...
+    smd.del_hook()
```
jarednielsen: How do these new tests differ from the existing tests that are already in the codebase? Might be better to just update the existing Keras tests.

vandanavk: not very different. its a mix of existing tests. json config from tf 1.x tests and model from tf 2.x test. this is meant only for TF 2.x. we are testing eager and non-eager mode in TF 2.x.

tests/zero_code_change/tensorflow2_integration_tests.py
vandanavk: done in the latest commit

tests/zero_code_change/tensorflow2_integration_tests.py
```diff
@@ -45,28 +45,31 @@ def get_keras_data():
...
 
-        model.compile(
-            loss="sparse_categorical_crossentropy",
-            optimizer=tf.keras.optimizers.RMSprop(),
-            metrics=["accuracy"],
-        )
+        opt = tf.keras.optimizers.RMSprop()
         if script_mode:
             hook = smd.KerasHook(out_dir=sim.out_dir, export_tensorboard=True)
+            opt = hook.wrap_optimizer(opt)
```
jarednielsen: Was the test working without wrap_optimizer() before?

vandanavk: yes, it was.  vanilla TF 2.x doesn't call get_gradients() or add_Weights that are in wrap_optimizer. so its a basically a don't care. added it here to catch any possible changes around this in future versions of TF.

vandanavk:
Any other concerns you see @jarednielsen @rahul003 ?
Any other tests you would recommend adding?

smdebug/tensorflow/base_hook.py
```diff
@@ -420,7 +421,10 @@ def set_gradients(self, gradients=None, gradients_and_variables=None):
         # TF 2.x doesn't provide gradient/optimizer variable names and values by default.
         # Skipping set_gradients and set_optimizer_variables for Tf 2.x until there is
         # support to pass names and values from TF side.
-        if is_tf_version_2x() and tf.executing_eagerly():
+
+        # From TF 2.2, executing_eagerly_outside_functions() should be called using
+        # tf.compat.v1.executing_eagerly_outside_functions()
+        if is_tf_version_2x() and ops.executing_eagerly_outside_functions():
```
jarednielsen: Probably want to address the function name change now then - TF 2.2 will be released soon.

smdebug/tensorflow/base_hook.py
```diff
@@ -441,7 +445,10 @@ def set_optimizer_variables(self, optimizer_variables):
         # TF 2.x doesn't provide gradient/optimizer variable names and values by default.
         # Skipping set_gradients and set_optimizer_variables for Tf 2.x until there is
         # support to pass names and values from TF side.
-        if is_tf_version_2x() and tf.executing_eagerly():
+
+        # From TF 2.2, executing_eagerly_outside_functions() should be called using
+        # tf.compat.v1.executing_eagerly_outside_functions()
```
jarednielsen: Likewise

tests/tensorflow2/test_keras.py
```diff
@@ -238,3 +242,57 @@ def test_weights_collections(out_dir, tf_eager_mode):
...
+    assert len(trial.tensor_names(collection=CollectionKeys.BIASES)) == 2
+    assert len(trial.tensor_names(collection=CollectionKeys.WEIGHTS)) == 2
+    assert len(trial.tensor_names(collection=CollectionKeys.LOSSES)) == 1
+    assert len(trial.tensor_names(collection=CollectionKeys.METRICS)) == 3
+
+
+@pytest.mark.slow
+def test_hook_from_json(tf_eager_mode, monkeypatch):
+    out_dir = "/tmp/test"
+    shutil.rmtree(out_dir, ignore_errors=True)
```
jarednielsen: Might want to improve the cleanup logic here - use a TemporaryDirectory context manager? https://docs.python.org/3/library/tempfile.html . This test will leave files left over.

tests/zero_code_change/tensorflow2_integration_tests.py
```diff
@@ -45,28 +45,31 @@ def get_keras_data():
     return (x_train, y_train), (x_test, y_test)
 
 
-def test_keras_v2(script_mode: bool = False, eager_mode: bool = True):
-    """ Works as intended. """
+def helper_test_keras_v2(script_mode: bool = False, eager_mode: bool = True):
+    """ Test the deafult ZCC behavior of saving losses and metrics in eager and non-eager modes."""
```
jarednielsen: Typo in deafult

smdebug/tensorflow/base_hook.py
vandanavk: this release will go into TF 2.1 container which wont have this name. so we'll have to make the change for the next container

vandanavk: it fails now if i use tf.compat.v1

tests/zero_code_change/tensorflow2_integration_tests.py
```diff
@@ -88,6 +191,11 @@ def test_keras_v2(script_mode: bool = False, eager_mode: bool = True):
...
-    test_keras_v2(script_mode=script_mode)
+    test_keras_v2_default(script_mode)
+    test_keras_v2_multi_collections(script_mode)
+    test_keras_v2_save_all(script_mode)
+
     # non-eager mode
-    test_keras_v2(script_mode=script_mode, eager_mode=False)
+    test_keras_v2_default(script_mode, eager_mode=False)
+    test_keras_v2_multi_collections(script_mode, eager_mode=False)
+    test_keras_v2_save_all(script_mode, eager_mode=False)
```
jarednielsen: Might be able to trim down the test logic by just testing the save_all case. There are unit tests to ensure that collections work properly, this file should really just focus on the ZCC aspects.

smdebug/tensorflow/base_hook.py
jarednielsen: smdebug should work for TF 2.1 and 2.2 simultaneously though, it's not just made for a specific container. Might need to wrap in a try/except or version check here.

tests/zero_code_change/tensorflow2_integration_tests.py
vandanavk: the multi collections case was the one that failed on the previous PR. so explicitly added a test for that case. might be redundant but thought to have it in there for now. what do you think?

tests/zero_code_change/tensorflow2_integration_tests.py
vandanavk: modified in the latest commit

tests/tensorflow2/test_keras.py
vandanavk: done in the latest commit

smdebug/tensorflow/base_hook.py
vandanavk: this release will go into TF 2.1 container which wont have this name. so we'll have to make the change for the next container. it fails now if i use tf.compat.v1

smdebug/tensorflow/base_hook.py
vandanavk: just verified with 2.2.0-rc0. there is no error when ops.executing_eagerly_outside_functions()
is used. 
in 2.2, both tf.compat.v1.executing_eagerly_outside_functions() and ops.executing_eagerly_outside_functions() will work.
in 2.1, only ops.executing_eagerly_outside_functions() will work. using tf.compat.v1.executing_eagerly_outside_functions() throws AttributeError exception

vandanavk: will change the comment here

vandanavk:
All review comments address @jarednielsen

anirudhacharya:
It might be better if this PR was pointed to the 0.7 release branch, instead of the master. The extra commit can later be cherry picked back into the master. 

Just a suggestion, to prevent extra/unnecessary changes to go into the 0.7.1 release.

PR #186, opened 2020-03-21
Experimental support for TF 2.x GradientTape

vandanavk:
### Description of changes:
This PR introduces new APIs to support TF 2.x GradientTape in script mode (user has to make certain calls to save different tensors).
Using GradientTape in a custom training loop, gives smdebug the option to save gradients, weights, biases, losses, metrics (Using fit(), only weights, biases, losses, metrics could be saved).

Also, with GradientTape, gradients, loss and metrics are EagerTensors and can be directly written to file. Tensor ref need not be used when the training script uses GradientTape.

The following implementations were considered before choosing option # 2:
1. Extending wrap_optimizer to wrap_apply_gradients:
With TF 2.x, the APIs get_gradients() and get_weights() in wrap_optimizer were not called as they were not used by Tensorflow in eager mode. To support GradientTape, one option explored was wrap apply_gradients call which could provide gradients and variables (weights, biases). Since apply_gradients is a common call between TF 1.x, TF 2.x eager and non-eager modes, TF 2.x fit() and GradientTape() approaches, it is hard to distinguish and use wrap apply gradients only for TF 2.x GradientTape()

2. Introducing wrap_tape():
GradientTape.gradient() API provides all the values that are required (loss, weight, bias, grads) with one API. Similar, to wrap_optimizer(), we could have the user make a call to wrap_tape() where we get values from gradient(). This would require changes from the normal way of writing a GradientTape code block.

Normal training script:
```
with tf.GradientTape(persistent=True) as tape:
    logits = model(data, training=True)  # (32,10)
    loss_value = cce(labels, logits)
```

If we introduced wrap_tape:
```
with hook.wrap_tape(tf.GradientTape(persistent=True)) as tape:
    logits = model(data, training=True)  # (32,10)
    loss_value = cce(labels, logits)
```

3. Introducing new APIs like record_tensor_value(), forward_pre_hook(), save_grads_vars(), forward_post_hook()
TF 2.x doesn't provide hooks like PT and MX. Introducing these APIs similar to PT hook. The idea is that ZCC will also call into these APIs as applicable.

4. Introduce hooks similar to PT and MX
This can be done in ZCC for eager execution. This still will not apply when execution is within a tf_function. Option# 3 may be required for vanilla TF

5. Generic save_tensor() API
Introducing a save_tensor() PI which could write any type of tensor was considered, but it was not the popular option because 
- introduces the possibility of multiple edge cases of when this API can be called
- do we cache all calls like we cache in the case of scalars?

6. Create a new smdebug hook called GradientTapeHook as a subclass of KerasHook which resembles PT or MX hooks. The APIs from option 3 would move to this new hook. This would ensure that fit() API users don't use GradientTape specific APIs. But this will require possibly duplicating code or re-write for handling the distributed case.

### Tensors saved (saveall):

['accuracy',
'gradients/dense/biasGrad',
'gradients/dense/kernelGrad',
'gradients/dense_1/biasGrad',
'gradients/dense_1/kernelGrad',
'loss',
'weights/dense/bias:0',
'weights/dense/kernel:0',
'weights/dense_1/bias:0',
'weights/dense_1/kernel:0']

#### Style and formatting:


smdebug/tensorflow/base_hook.py
```diff
@@ -180,7 +180,8 @@ def _get_default_collections(self):
         return DEFAULT_TF_COLLECTIONS
 
     def export_collections(self):
-        assert self._prepared_tensors[self.mode]
+        if not self.tape:
```
jarednielsen: You'll need to define the `tape` attribute in the base class, rather than KerasHook, if methods here are referring to it. This will cause an AssertionError if the method is called on a SessionHook where `tape` is not defined.

smdebug/tensorflow/keras.py
```diff
@@ -620,3 +621,75 @@ def new_add_weight(opt, *args, **kwargs):
         # Optimizer is being saved to support additional features in the future.
         self.optimizer = optimizer
         return optimizer
+
+    def forward_pre_hook(self, tape):
+        self.tape = tape
+        self.set_mode(ModeKeys.TRAIN)
```
jarednielsen: How do we know that it's training mode? Forward pass also occurs in evaluation.

smdebug/tensorflow/keras.py
```diff
@@ -620,3 +621,75 @@ def new_add_weight(opt, *args, **kwargs):
...
+        if self._is_not_supported():
+            return
```
jarednielsen: Log an warning here?

smdebug/tensorflow/keras.py
```diff
@@ -620,3 +621,75 @@ def new_add_weight(opt, *args, **kwargs):
...
+            # will only have tensor names from first mode
+
+            # this means sometimes collections will be exported after 1 step
+            self.export_collections()
+            self._exported_collections = True
+
+    def update_step(self, grads, vars):
+        if self._get_collections_to_save_for_step():
+            if grads is not None and vars is not None:
+                for idx, (g, v) in enumerate(zip(grads, vars)):
```
jarednielsen: Nit: I don't see why enumerate is used here.

smdebug/tensorflow/keras.py
```diff
@@ -620,3 +621,75 @@ def new_add_weight(opt, *args, **kwargs):
...
+    def record_tensor_value(self, tensor_name, tensor_value, tape):
+        self.tape = tape
+
+        def add_metric_to_coll(self, tensor_name, tensor_value):
+            if tensor_name not in self.tensor_to_collections:
+                matched_colls = set()
+                if tensor_name in ["loss", "val_loss"]:
+                    coll_name = CollectionKeys.LOSSES
+                else:
+                    coll_name = CollectionKeys.METRICS
```
jarednielsen: Seems like this logic should be covered by the CollectionKey regexes. Can you elaborate why we're hardcoding this instead of referring to existing collections methods?

smdebug/tensorflow/collection.py
```diff
@@ -114,6 +114,11 @@ def set_tensor_ref(self, tensor):
         self._tensors[tensor.name] = tensor
         self.add_tensor_name(tensor.export_name)
 
+    def set_tensor(self, tensor_name, tensor):
```
jarednielsen: More descriptive method name and docstring - unclear what is being set.

smdebug/tensorflow/keras.py
```diff
@@ -55,6 +55,7 @@ def __init__(
         self.tensor_refs_to_save_this_step = set()
         self._fetches_added = set()
         self.callable_cache = CallableCache()
+        self.tape = None
```
jarednielsen: Since this variable is referred to in the base hook, it should be set in the base hook constructor. Otherwise the base hook method should be moved out to this class.

smdebug/tensorflow/keras.py
```diff
@@ -620,3 +621,75 @@ def new_add_weight(opt, *args, **kwargs):
...
+
+        if self._exported_collections is False:
+            # in keras, these collections change when mode changes
+            # but rest of the project isn't yet capable of handling this
+            # this means that collections like outputs, or other collections with intermediate tensors
+            # will only have tensor names from first mode
+
+            # this means sometimes collections will be exported after 1 step
+            self.export_collections()
+            self._exported_collections = True
```
jarednielsen: This seems like a pretty important gotcha when users run both training and validation each epoch. Let's brainstorm more how to improve the customer experience in this case, don't want users to be confused.

smdebug/tensorflow/collection.py
```diff
@@ -133,7 +138,10 @@ def __init__(self, collections=None, create_default=True):
         if create_default:
             for n in DEFAULT_TF_COLLECTIONS:
                 self.create_collection(n)
-            self.get(CollectionKeys.BIASES).include("bias")
+            self.get(CollectionKeys.BIASES).include("^(?!gradient).*bias")
+            self.get(CollectionKeys.WEIGHTS).include("^weights/.*/((?!bias).)*$")
```
jarednielsen: My regex-fu is limited, but looks like this will exclude bias gradients in CollectionKeys.BIASES, but include weight gradients in CollectionKeys.WEIGHTS. Is that what we want?

tests/tensorflow2/test_keras_gradtape.py
```diff
@@ -0,0 +1,252 @@
+# Third Party
```
jarednielsen: Add docstrings to all tests describing expected behavior and what is being tested. It's tough to follow what's going on with all the test parameters and conditional paths.

jarednielsen:
Since we're introducing a new API, updating the documentation as part of this PR would be very helpful.

smdebug/tensorflow/keras.py
vandanavk: there's a regex for Loss but i dont think there's a regex for metrics

smdebug/tensorflow/collection.py
vandanavk: its set_tensor_ref but without tensor_Ref usage.
just assign the tensor as is

smdebug/tensorflow/collection.py
vandanavk: weight gradients are called "gradients/dense/kernelGrad".
bias gradients are called "gradients/dense/biasGrad". hence, only bias needs a special regex. similar to PT

smdebug/tensorflow/keras.py
vandanavk: Agree, will rely on hook.set_mode called from the training script

smdebug/tensorflow/keras.py
vandanavk: There are logs inside _is_not_supported()

smdebug/tensorflow/keras.py
```diff
@@ -620,3 +623,86 @@ def new_add_weight(opt, *args, **kwargs):
...
+            # as user's code after hook might add collections
+            self._prepare_collections()
+            self.prepared_collections = True
+
+        self._increment_step()
+
+        if self._get_collections_to_save_for_step():
+            self._initialize_writers()
+
+    def set_grads_vars(self, grads, vars):
```
vandanavk: We could modify base_hook.set_gradients() to handle this case instead of introducing a new API.

vandanavk:
@jarednielsen ready for another round of review.
Will update the docs once the APIs are finalized.
The CI tests are failing because test_keras.py is executed first - eager and then non-eager mode, followed by test_keras_gradtape.py which continues in non-eager mode is what I suspect. Looking into how to fix this. Executing test_keras_gradtape.py just by itself is successful.

smdebug/tensorflow/keras.py
```diff
@@ -337,7 +337,7 @@ def _save_inputs(self, check_before_write=True):
         # TODO
         pass
 
-    def _add_metric(self, metric_name):
+    def _add_metric(self, metric_name, metric_value=None):
```
jarednielsen: Adding type annotations makes it clear whether a bool/str/tensor is expected.

jarednielsen:
> @jarednielsen ready for another round of review.
> Will update the docs once the APIs are finalized.
> The CI tests are failing because test_keras.py is executed first - eager and then non-eager mode, followed by test_keras_gradtape.py which continues in non-eager mode is what I suspect. Looking into how to fix this. Executing test_keras_gradtape.py just by itself is successful.

Writing the docs first will help with API design, so I'd recommend putting a documentation draft together before finalizing the design. It would also make it clearer whether we're using option 1, 2, 3, 4, 5, or some combination of them.

smdebug/tensorflow/keras.py
jarednielsen: Might be worth adding one then. There's significant codebase complexity that comes from passing around collections, seeing if collections are prepared, and seeing if tensors match the collection - duplicated all over the place. The more we can refactor and simplify, the better :)

smdebug/tensorflow/keras.py
vandanavk: done in the latest commit

vandanavk:
> > @jarednielsen ready for another round of review.
> > Will update the docs once the APIs are finalized.
> > The CI tests are failing because test_keras.py is executed first - eager and then non-eager mode, followed by test_keras_gradtape.py which continues in non-eager mode is what I suspect. Looking into how to fix this. Executing test_keras_gradtape.py just by itself is successful.
> 
> Writing the docs first will help with API design, so I'd recommend putting a documentation draft together before finalizing the design. It would also make it clearer whether we're using option 1, 2, 3, 4, 5, or some combination of them.

thanks @jarednielsen. updated the docs in the latest commit. Also added an option 6 to the PR description :)

smdebug/tensorflow/keras.py
vandanavk: let me try, this might require enforcing that the metric name includes a "metric" word. 

reused the existing add_metric() in Keras hook for now.

tests/tensorflow2/test_keras_gradtape.py
vandanavk: done in the more recent commits. 

also, had to move up gradtape test to run before fit() tests. there's something going on in the fit() execution that makes the gradtape tests fail if run after fit(). I've tried `tf.reset_default_graph()`, `tf.keras.backend.clear_session()` (FWIW), `tf.enable_eager_execution()` but none worked.

vandanavk:
Based on offline discussion with Jared, trying out option 2. wrap_tape() based implementation.

smdebug/tensorflow/keras.py
```diff
@@ -620,3 +623,129 @@ def new_add_weight(opt, *args, **kwargs):
...
+        def unwrap(func):
+            while _is_wrapper(func):
+                func = func.__wrapped__
+            return func
+
+        self.tape.__class__._push_tape = unwrap(self.tape.__class__._push_tape)
+        self.tape.__class__._pop_tape = unwrap(self.tape.__class__._pop_tape)
+        self.tape.__class__.gradient = unwrap(self.tape.__class__.gradient)
+
+    def close(self):
```
vandanavk: @jarednielsen Adding this in close() will require the user to call close() for sure.
Do you think unwrap_tape is more suitable in _cleanup(self) ?

config/tests.sh
```diff
@@ -29,7 +29,12 @@ run_for_framework() {
       fi
 
     else
-      python -m pytest --durations=50 --html=$REPORT_DIR/report_$1.html -v -s --self-contained-html tests/$1
+      if [ "$1" = "tensorflow2" ] ; then
+        python -m pytest --durations=50 --html=$REPORT_DIR/report_$1/eager_mode.html -v -s --self-contained-html tests/$1
+        python -m pytest --durations=50 --non-eager --html=$REPORT_DIR/report_$1/non_eager_mode.html -v -s --self-contained-html tests/$1
```
jarednielsen: Relying on test execution order can be flaky. Better to reset things in the test code so tests can truly be run independently.

tests/zero_code_change/tensorflow2_integration_tests.py
```diff
@@ -157,26 +157,18 @@ def helper_test_keras_v2_gradienttape(script_mode: bool = False, json_file_conte
...
+            tape = hook.wrap_tape(tape)
             for epoch in range(n_epochs):
                 for data, labels in dataset:
                     dataset_labels = labels
                     labels = tf.one_hot(labels, depth=10)
-                    with tf.GradientTape(persistent=True) as tape:
-                        # call this API before a forward pass to prepare collections
-                        # and initialize writers
-                        hook.forward_pre_hook(tape)
+                    with tape:
```
jarednielsen: Nit: Encouraging `with hook.wrap_tape(tf.GradientTape())` rather than declaring variables in advance is even cleaner, then the user only has a one-line change to make.

smdebug/tensorflow/keras.py
jarednielsen: Yes

smdebug/tensorflow/keras.py
```diff
@@ -626,97 +626,157 @@ def new_add_weight(opt, *args, **kwargs):
...
-        Export collections and model at the end of a step.
+            if self._exported_collections is False:
+                # in keras, these collections change when mode changes
+                # but rest of the project isn't yet capable of handling this
+                # this means that collections like outputs, or other collections with intermediate tensors
+                # will only have tensor names from first mode
+
+                # this means sometimes collections will be exported after 1 step
+                self.export_collections()
+                self._exported_collections = True
```
jarednielsen: Some hooks use the variable `self.exported_collections` and some use `self._exported_collections`. Maybe not within the scope of this PR, but that should be fixed, and setting the marker boolean to True should occur within the method, not outside of it.

tests/tensorflow2/test_keras.py
```diff
@@ -471,11 +465,10 @@ def test_include_regex(out_dir, tf_eager_mode):
         assert tr.tensor(tname).value(0) is not None
 
 
+@pytest.mark.skip_if_non_eager
```
jarednielsen: Let's see what we can do so all tests can be run at once, without relying on test scaffolding to preserve the environment

tests/tensorflow2/test_keras.py
```diff
@@ -361,11 +356,10 @@ def test_gradtape_hook_from_json(out_dir, monkeypatch):
 
     trial = smd.create_trial(path=out_dir)
     # can't save gradients in TF 2.x
-    assert len(trial.tensor_names()) == 4
+    assert len(trial.tensor_names()) == 3  # 4
```
jarednielsen: What's happening to metrics? Looks like those are no longer being recorded.

tests/tensorflow2/test_keras.py
```diff
@@ -155,6 +155,9 @@ def helper_keras_gradtape(
 
     opt = hook.wrap_optimizer(opt)
 
+    tape = tf.GradientTape(persistent=True)
+    tape = hook.wrap_tape(tape)
+
```
jarednielsen: See comment above about making this a one-liner.

tests/tensorflow2/test_keras.py
vandanavk: Metrics wasn't getting exported as export_collection was called in pop_tape() before user could manually save metrics tensor.
Modified the code to behave similar to PT hook - export_collections in push_tape and introduce record_tensor_value

tests/tensorflow2/test_keras.py
vandanavk: done in the latest commit

config/tests.sh
vandanavk: Since we can't toggle between eager and non-eager execution, introduced this change. https://github.com/awslabs/sagemaker-debugger/pull/190/files#diff-484462fced51d1a06b1d93b4a44dd535R54.

Is it possible to reset? I couldn't much about this scenario in the tensorflow documentation or GitHub issue list.

This comment also gave https://github.com/awslabs/sagemaker-debugger/blob/989a4d15e17541b06027d6f658f1fd15fab63566/tests/tensorflow/keras/test_keras.py#L137 also gave me the impression that previous attempts were unsuccessful.

tests/zero_code_change/tensorflow2_integration_tests.py
vandanavk: done in the latest commit

config/tests.sh
vandanavk: also, this change is in this PR and #190 . will remove it from either depending on which gets merged first

tests/tensorflow2/test_keras.py
jarednielsen: Is persistent=True still necessary?

vandanavk: not necessary (by default persistent=False). changing it in the latest commit.

added a test for persistent=True and made a few changes to handle multiple calls to tape.gradient in this case.

smdebug/tensorflow/keras.py
vandanavk: done in the latest commit

docs/api.md
```diff
@@ -156,6 +156,8 @@ Note that there are three types of Hooks in TensorFlow: SessionHook, EstimatorHo
...
 | `add_to_collection(`<br/>  `collection_name, variable)` | `collection_name (str)` : name of the collection to add to. <br/> `variable` parameter to pass to the collection's `add` method. | `None` | Calls the `add` method of a collection object. See [this section](#collection) for more. |
+| `wrap_tape(tape)` | `tape` (tensorflow.python.eager.backprop.GradientTape) | Returns a tape object with three identifying markers to help `smdebug`. This returned tape should be used for training. | When not using Zero Script Change environments, calling this method on your tape is necessary for SageMaker Debugger to identify and save gradient tensors. Note that this method returns the same tape object passed.
```
Vikas-kum: are these specific to TF2.x ?
I didn't get what is special in record_tensor_value that makes is only for TF2.x

docs/tensorflow.md
```diff
@@ -74,6 +74,26 @@ model.fit(x_train, y_train, epochs=args.epochs, callbacks=[hook])
...
+```python
+import smdebug.tensorflow as smd
+hook = smd.KerasHook(out_dir=args.out_dir)
+
+model = tf.keras.models.Sequential([ ... ])
+    for epoch in range(n_epochs):
+        for data, labels in dataset:
+            dataset_labels = labels
+            # wrap the tape to capture tensors
+            with hook.wrap_tape(tf.GradientTape(persistent=True)) as tape:
```
Vikas-kum: Is there a zero code change behavior for this?

Vikas-kum: will this work if persistent=False?

docs/tensorflow.md
```diff
@@ -74,6 +74,26 @@ model.fit(x_train, y_train, epochs=args.epochs, callbacks=[hook])
...
+                logits = model(data, training=True)  # (32,10)
+                loss_value = cce(labels, logits)
+            grads = tape.gradient(loss_value, model.variables)
+            opt.apply_gradients(zip(grads, model.variables))
+            acc = train_acc_metric(dataset_labels, logits)
+            # manually save metric values
+            hook.record_tensor_value(tensor_name="accuracy", tensor_value=acc)
```
Vikas-kum: Do we need to explicitly call record_tensor_value even if I have wrapped the tape? Or this is only for recording arbitrary tensor ?

smdebug/tensorflow/base_hook.py
```diff
@@ -180,7 +183,8 @@ def _get_default_collections(self):
...
```
Vikas-kum: else ?

smdebug/tensorflow/base_hook.py
```diff
@@ -369,13 +373,18 @@ def _log_unsupported_optimizer(self, optimizer):
 
     def _get_collections_with_tensor(self, tf_tensor_name) -> Set["Collection"]:
         self._assert_prep()
+        if self.tape:
+            return super()._get_collections_with_tensor(tf_tensor_name)
```
Vikas-kum: why is this special one and why can't we put tf_tensor_name in tensor_to_collections map ?

docs/api.md
vandanavk: wrap_tape is specific to TF 2.x.

record_tensor_value() is generic, but restricted to metrics value.

docs/tensorflow.md
vandanavk: no zero code change behavior yet. will be introduced in the next release.

by default persistent=False (all the tests handle this case). plus there is one test for persistent=True

docs/tensorflow.md
vandanavk: this is only for metric tensor. we get loss, grads, weights, bias through wrap tape

smdebug/tensorflow/base_hook.py
vandanavk: there is no prepare_layers when GradientTape is used. all tensors to be saved are EagerTensors, so the value is available to be written right away.

This behavior is similar to PT and MX behavior

smdebug/tensorflow/base_hook.py
vandanavk: in general for TF, tensor_to_collections is populated in prepare_layers().

this is not applicable to GradientTape as the tensors are EagerTensors and can be written as soon as we get them through tape.gradient().

so in case of GradientTape, tensor_to_collections has to be populated through core/hook.py's _get_collections_with_tensor and there is a return of self.tensor_to_collections[tensor_name] there

docs/api.md
vandanavk: there is no regex for metrics. so by using this API, user can save any tensor but according to the check in this function, it will get saved as METRICS collection (if the user chooses to save metrics collection).

vandanavk: did not create a generic save_tensor API as that will be a feature by itself.

used record_tensor_value() similar to the usage in PT hook for functional loss

smdebug/tensorflow/base_hook.py
vandanavk: else, probably assert self.prepared_collections

PR #190, opened 2020-03-26
Support SyncOnReadVariable in mirrored strategy

vandanavk:
### Description of changes:
There is a new variable of type values.DistributedVariable called SyncOnReadVariable used for performance improvement in MirroredStrategy.
Users may face issues in case this new type of variable is used during distributed training.

Added tests for TF 2 keras fit with Mirrored strategy.

Also made some modifications to conftest and tests.sh. In TF, once we disable eager execution, we cannot re-enable eager execution. The modifications in these 2 files will ensure that all tests are run in eager mode first followed by non-eager mode.

#### Style and formatting:

### DIFF
--- a/smdebug/tensorflow/collection.py
+++ b/smdebug/tensorflow/collection.py
@@ -81,7 +81,7 @@ class Collection(BaseCollection):
             return self.add_variable(arg, mode=mode)
         elif isinstance(arg, tf.Tensor):
             return self.add_tensor(arg, mode=mode)
-        elif isinstance(arg, values.MirroredVariable):
+        elif isinstance(arg, (values.MirroredVariable, values.DistributedVariable)):
             return self.add_mirrored_variable(arg, mode=mode)
         elif isinstance(arg, values.AggregatingVariable):
             return self.add_aggregating_variable(arg, mode=mode)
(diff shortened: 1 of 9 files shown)


smdebug/tensorflow/collection.py
```diff
@@ -81,7 +81,7 @@ def add_for_mode(self, arg, mode=None):
             return self.add_variable(arg, mode=mode)
         elif isinstance(arg, tf.Tensor):
             return self.add_tensor(arg, mode=mode)
-        elif isinstance(arg, values.MirroredVariable):
+        elif isinstance(arg, (values.MirroredVariable, values.DistributedVariable)):
```
vandanavk: both MirroredVariable and SyncOnReadVariable are of type values.DistributedVariable. Probably could just check for values.DistributedVariable

NihalHarish: If both values.MirroredVariable and SyncOnReadVariable are of the type, then it probably makes sense to type check only the parent.

Also can you rename the `self.add_mirrored_variable`

vandanavk: done in the latest commit

NihalHarish:
Add comments explaining the changes you have made in the test suite.

PR #191, opened 2020-03-27
Unambiguous Initialization of Parent Classes

NihalHarish:
### Description of changes:
- encountered a bug while testing the tensorflow 2.2 release candidate, where the member variables of the parent class were unavailable from the child class.
- this code change will make the initialization of the parent classes unambiguous 

#### Style and formatting:


vandanavk:
SessionHook also inherits from 2 parents. Can you make the same change to the init call there?

NihalHarish:
SessionHook inherits from tf.train.SessionRunHook and TensorflowBaseHook, out of which only TensorflowBaseHook has an init function, so it is unambiguous.

Issue #184, opened 2020-03-19
Crash occurs when trying to register a hook on a tensor that doesn't require gradients


jbschlosser:
**Background**: I am trying to run a SageMaker Training Job with custom PyTorch 1.4.0 code using the official sagemaker-pytorch-container (763104351884.dkr.ecr.us-east-1.amazonaws.com/pytorch-training:1.4.0-gpu-py36-cu101-ubuntu16.04) on a single ml.p3.2xlarge instance.

**Issue**: I get the following error: `RuntimeError: cannot register a hook on a tensor that doesn't require gradient`. Issue is related to the fact that, in my code, I'm loading a previously trained `torch.nn.Embedding` that does not need gradient updates. Disabling SageMaker Debugger by passing `debugger_hook_config=False` to the Estimator constructor gets past the error, but then training hangs indefinitely after the first training step.

Vikas-kum:
@jbschlosser Thanks for reporting this. Would you be able to provide the script to reproduce this problem.

NihalHarish:
@jbschlosser 

I ran the example from https://github.com/pytorch/examples/tree/master/imagenet 
with 

`python main.py -a vgg16 --lr 0.01 --pretrained  --epochs 1 tiny-imagenet-200/`

using the docker 
`763104351884.dkr.ecr.us-east-1.amazonaws.com/pytorch-training:1.4.0-gpu-py36-cu101-ubuntu16.04`

and **smdebug version: 0.7.1**
And it did not throw the error you have reported, are there any changes I need to make to the script I have used to recreate the issue?

jbschlosser:
Thanks for the quick response! Here's a semi-minimal example I threw together that reproduces the error:

sm_train.py:
```python
import argparse
from sagemaker.pytorch import PyTorch


def do_sm_training(args):
    # Create the Estimator.
    estimator = PyTorch(
        base_job_name='train-debug-example',
        entry_point='train.py',
        role=args.role,
        train_instance_count=1,
        train_instance_type='ml.p3.2xlarge',
        output_path=args.output_path,
        framework_version='1.4.0',
        py_version='py3',
        tags=[]
    )
    # Train the Estimator.
    estimator.fit({
        'train': args.train
    }, wait=True)


if __name__ == '__main__':
    parser = argparse.ArgumentParser()
    parser.add_argument('--train', type=str, required=True,
                        help='Path to input data in S3')
    parser.add_argument('--role', type=str, required=True,
[...]

PR #193, opened 2020-03-27
Register If Tensor Requires Gradients

NihalHarish:
### Description of changes:
- The hook crashes with the error `RuntimeError: cannot register a hook on a tensor that doesn't require gradient` when training with pretrained embeddings, see Issue #184 for more details
- This PR adds a check to see if the tensor requires gradient before registering the hook

#### Style and formatting:

#184 **Crash occurs when trying to register a hook on a tensor that doesn't require gradients**

### DIFF
--- a/tests/zero_code_change/test_training_with_no_grad_updates.py
+++ b/tests/zero_code_change/test_training_with_no_grad_updates.py
@@ -104,7 +104,7 @@ def do_training():
         optimizer.step()
 
 
-if __name__ == "__main__":
+def test_training_with_no_grad_updates():
     temp_dir = TemporaryDirectory().name
     path = build_json(temp_dir, include_collections=["losses"], save_interval="1")
     os.environ["SMDEBUG_CONFIG_FILE_PATH"] = str(path)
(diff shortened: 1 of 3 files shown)


tests/zero_code_change/test_training_with_no_grad_updates.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # Setup dataset + data loader.
+    BATCH_SIZE = 64
+    NUM_BATCHES = 100
+    train_dataset = TestDataset(num_items=N, num_samples=NUM_BATCHES * BATCH_SIZE)
+    train_loader = DataLoader(train_dataset, batch_size=BATCH_SIZE, shuffle=False)
+
+    # Setup model.
+    model = TestModel(embedding_size=EMBEDDING_SIZE).to(get_device())
+
```
NRauschmayr: I think you need to set requires_grad to False, if the testcase is intended to check hook on parameters that do not require gradients. I am getting the following output:
```model = TestModel(10)
for parameter in model.parameters():
    print(parameter.requires_grad)
True
True
```

vandanavk:
Can you move the test into pytorch_integration_tests.py ? All PT ZCC tests are there. Also, by adding a new file in this folder, you may have to update various test and release pipelines to execute this test.

PR #198, opened 2020-04-04
Version Bump

NihalHarish:
### Description of changes:
- Version bump 0.8

#### Style and formatting:


NihalHarish:
> Earlier it was not starting with v
> Why v now?



According to semver, version strings **should not** have a prefixed 'v', so I will **revert** this.
https://semver.org/spec/v2.0.0.html#is-v123-a-semantic-version

But, we currently use the same name for both the branch and tag, which should be avoided.
Maybe we can have branch names contain only MAJOR.MINOR [0.7, 0.8, ...] or we can append 'v' to tag names (https://semver.org/#is-v123-a-semantic-version) and change the DLC code to append 'v' to the version number to match the correct tag.

PR #200, opened 2020-04-06
Bug Fix to prevent repeated writes to metadata.json

anirudhacharya:
### Description of changes:
During a Sagemaker training job, there was an issue of the `metadata.json` file in the `checkpoints_dir` being re-written every second.

This probably happened because we use `json.dumps` [here](https://github.com/awslabs/sagemaker-debugger/blob/master/smdebug/core/state_store.py#L104) to write to the `states_file` every time the `last_modified_time` of the `checkpoint_dir` is greater than `last_Saved time`( [checked here](https://github.com/awslabs/sagemaker-debugger/blob/master/smdebug/core/state_store.py#L75)). 

`json.dumps` may not guarantee that file was actually written to disk. The file could be in cache and flushed to disk asynchronously a second later. In such a case the `last_modified_time` of the file will always be more than `last_Saved time`. 

To fix it I have filtered out the `states_file` while determining if the `checkpoint_dir` has been updated( [here](https://github.com/awslabs/sagemaker-debugger/blob/master/smdebug/core/state_store.py#L82)).

#### Style and formatting:

### DIFF
--- a/config/buildspec_zero_code_change.yml
+++ b/config/buildspec_zero_code_change.yml
@@ -31,8 +31,8 @@ phases:
 
   build:
     commands:
-      - cd $CODEBUILD_SRC_DIR  && chmod +x config/tests.sh && PYTHONPATH=. && ./config/tests.sh && cd ..
-      - cd $CODEBUILD_SRC_DIR_RULES && chmod +x config/tests.sh && PYTHONPATH=. && ./config/tests.sh  && cd ..
+      - cd $CODEBUILD_SRC_DIR  && chmod +x config/tests.sh && PYTHONPATH=. ./config/tests.sh && cd ..
+      - cd $CODEBUILD_SRC_DIR_RULES && chmod +x config/tests.sh && PYTHONPATH=. ./config/tests.sh  && cd ..
 
   post_build:
     commands:
--- a/config/buildspec_zero_code_change.yml
+++ b/config/buildspec_zero_code_change.yml
@@ -31,8 +31,8 @@ phases:
 
   build:
     commands:
-      - cd $CODEBUILD_SRC_DIR  && chmod +x config/tests.sh && PYTHONPATH=. ./config/tests.sh && cd ..
-      - cd $CODEBUILD_SRC_DIR_RULES && chmod +x config/tests.sh && PYTHONPATH=. ./config/tests.sh  && cd ..
+      - cd $CODEBUILD_SRC_DIR  && chmod +x config/tests.sh && PYTHONPATH=. && ./config/tests.sh && cd ..
+      - cd $CODEBUILD_SRC_DIR_RULES && chmod +x config/tests.sh && PYTHONPATH=. && ./config/tests.sh  && cd ..
 
   post_build:
     commands:
--- a/smdebug/core/state_store.py
+++ b/smdebug/core/state_store.py
@@ -74,14 +74,24 @@ class StateStore:
 
     def is_checkpoint_updated(self):
         """
-        Check whether the folder in which checkpoints are stored got updated.
-        Update to that folder indicates, user attempted to store the new checkpoint in that directory.
+        Check whether new checkpoint files got added or existing checkpoint files that are
+        stored got updated.
         """
         if self._checkpoint_dir is not None:
-            checkpoint_timestamp = max(
-                os.path.getmtime(child) for child, _, _ in os.walk(self._checkpoint_dir)
-            )
-            if checkpoint_timestamp > self._checkpoint_update_timestamp:
+            timestamps = [
+                os.path.getmtime(child)
+                for child, _, _ in os.walk(self._checkpoint_dir)
+                if child != self._checkpoint_dir
+            ]
+            _, _, files = next(os.walk(self._checkpoint_dir))
+            timestamps += [
+                os.path.getmtime(os.path.join(self._checkpoint_dir, file))
+                for file in files
+                if file != METADATA_FILENAME
+            ]
+            if not timestamps:
+                return False
+            if max(timestamps) > self._checkpoint_update_timestamp:
                 return True
         return False
 
--- a/smdebug/core/state_store.py
+++ b/smdebug/core/state_store.py
@@ -1,7 +1,6 @@
 # Standard Library
 import json
 import os
-import time
 
 # First Party
 from smdebug.core.config_constants import (
@@ -119,4 +118,4 @@ class StateStore:
         self._saved_states.append(ts_state)
         with open(self._states_file, "w") as out_file:
             json.dump(self._saved_states, out_file)
-        self._checkpoint_update_timestamp = time.time()
+        self._checkpoint_update_timestamp = os.path.getmtime(self._states_file)
--- a/smdebug/core/state_store.py
+++ b/smdebug/core/state_store.py
@@ -86,6 +86,12 @@ class StateStore:
             if not checkpoint_files:
                 return False
             timestamps = [os.path.getmtime(file) for file in checkpoint_files]
+            logger.debug(
+                f"Timestamps of different checkpoint files {[i for i in zip(checkpoint_files, timestamps)]}"
+            )
+            logger.debug(
+                f"Timestamp of the last checkpoint update: {self._checkpoint_update_timestamp}"
+            )
             if max(timestamps) > self._checkpoint_update_timestamp:
                 return True
         return False
--- a/smdebug/core/state_store.py
+++ b/smdebug/core/state_store.py
@@ -28,6 +28,7 @@ class StateStore:
         self._states_file = None
         self._checkpoint_dir = None
         self._retrieve_path_to_checkpoint()
+        self._max_timestamp_seen = 0
         if self._checkpoint_dir is not None:
             self._states_file = os.path.join(self._checkpoint_dir, METADATA_FILENAME)
             self._read_states_file()
@@ -94,9 +95,10 @@ class StateStore:
             logger.debug(
                 f"Timestamp of the last checkpoint update: {self._checkpoint_update_timestamp}"
             )
-            if max(timestamps) > self._checkpoint_update_timestamp:
+            self._max_timestamp_seen = max(timestamps)
+            if  max_timestamp_seen > self._checkpoint_update_timestamp:
                 logger.debug(
-                    f"The most recent timestamp of the checkpoint files: {max(timestamps)}"
+                    f"The most recent timestamp of the checkpoint files: {self._max_timestamp_seen}"
                 )
                 return True
         return False
@@ -118,4 +120,4 @@ class StateStore:
         self._saved_states.append(ts_state)
         with open(self._states_file, "w") as out_file:
             json.dump(self._saved_states, out_file)
-        self._checkpoint_update_timestamp = os.path.getmtime(self._states_file)
+        self._checkpoint_update_timestamp = self._max_timestamp_seen
--- a/smdebug/core/state_store.py
+++ b/smdebug/core/state_store.py
@@ -78,19 +78,14 @@ class StateStore:
         stored got updated.
         """
         if self._checkpoint_dir is not None:
-            timestamps = [
-                os.path.getmtime(child)
-                for child, _, _ in os.walk(self._checkpoint_dir)
-                if child != self._checkpoint_dir
-            ]
-            _, _, files = next(os.walk(self._checkpoint_dir))
-            timestamps += [
-                os.path.getmtime(os.path.join(self._checkpoint_dir, file))
-                for file in files
-                if file != METADATA_FILENAME
-            ]
-            if not timestamps:
+            checkpoint_files = []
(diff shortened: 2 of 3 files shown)


anirudhacharya:
There are a few issues with this fix that I have made. 
1. By modifying `is_checkpoint_updated` we have changed the nature of the function. By filtering out the `metadata.json` file we have  made the function inaccurate. So now even if the `checkpoint_dir` is modified `is_checkpoint_updated` function can still return `False`.

For example, the checkpoint_dir is like this -
```bash
checkpoint_dir
 |-- sub-dir1
 |-- sub-dir2
 |-- sub-dir3
 |-- metadata.json
```
If `metadata.json` is modified then it technically means that `checkpoint_dir` is also modified, so `is_checkpoint_updated` should return `True`. But the current modification to the function will not do that.

One way to fix the bug without modifying the behavior of the `is_checkpoint_updated` function might be to ensure that the write call to write the `states_file` is a synchronous call( [here](https://github.com/awslabs/sagemaker-debugger/blob/master/smdebug/core/state_store.py#L104))

@Vikas-kum kindly review this.

smdebug/core/state_store.py
```diff
@@ -79,7 +79,9 @@ def is_checkpoint_updated(self):
         """
         if self._checkpoint_dir is not None:
             checkpoint_timestamp = max(
-                os.path.getmtime(child) for child, _, _ in os.walk(self._checkpoint_dir)
+                os.path.getmtime(child)
+                for child, _, _ in os.walk(self._checkpoint_dir)
+                if child != os.path.dirname(self._states_file)
```
Vikas-kum, line 84: why dirname? You need to check for exact filename and filter out meradata.json 
Also write a test case for this.

smdebug/core/state_store.py
```diff
@@ -74,14 +74,24 @@ def _read_states_file(self):
...
+                for child, _, _ in os.walk(self._checkpoint_dir)
+                if child != self._checkpoint_dir
+            ]
+            _, _, files = next(os.walk(self._checkpoint_dir))
+            timestamps += [
+                os.path.getmtime(os.path.join(self._checkpoint_dir, file))
+                for file in files
+                if file != METADATA_FILENAME
+            ]
+            if not timestamps:
```
Vikas-kum, line 92: I think we need a simple way to recursively get all the files.  
import os
>>> for child, r, c in os.walk("/tmp/newlogsRunTest/"):
...   for f in c:
...     if f != "haha":
...       print(child,"/",f)

Also see here- 
https://stackoverflow.com/questions/2186525/how-to-use-glob-to-find-files-recursively

smdebug/core/state_store.py
```diff
@@ -74,14 +74,24 @@ def _read_states_file(self):
...
+                return False
+            if max(timestamps) > self._checkpoint_update_timestamp:
```
Vikas-kum, line 94: before returning we should also add logger debug and dump files considered anf their gmtime

tests/core/test_state_store.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def test_is_checkpoint_updated():
+    s1 = StateStore()
+    s1._checkpoint_update_timestamp = time.time()
+    s1._checkpoint_dir = setup_test1()
+    # the checkpoint update time is greater than _checkpoint_update_timestamp. is_checkpoint_updated should return true.
+    assert s1.is_checkpoint_updated()
+
+    s1._checkpoint_update_timestamp = time.time()
```
Vikas-kum, line 49: why this i set in test case. 

I think you require test case like:
test1:

State_store - check_updated() ?  returns False
create dir/files 
State_store - check_updated() ?  returns True 
update_state() 
state_store -> check_updated ? reutrns False 

update one of existing file : 
state_store -> check_updated ? returns True 
update_state()
Create _one_ new file 
state_store -> check_updated ? returns True

anirudhacharya:
> Also, please look at build failures.

this test `test_is_checkpoint_updated` which I added in this PR, failed. Surprising, since all these tests passed locally. I will see, what might have gone wrong.

anirudhacharya:
Three tf2 job will pass after I rebase on top of PR #203

config/buildspec_zero_code_change.yml
```diff
@@ -31,8 +31,8 @@ phases:
 
   build:
     commands:
-      - cd $CODEBUILD_SRC_DIR  && chmod +x config/tests.sh && PYTHONPATH=. ./config/tests.sh && cd ..
-      - cd $CODEBUILD_SRC_DIR_RULES && chmod +x config/tests.sh && PYTHONPATH=. ./config/tests.sh  && cd ..
+      - cd $CODEBUILD_SRC_DIR  && chmod +x config/tests.sh && PYTHONPATH=. && ./config/tests.sh && cd ..
```
Vikas-kum, line 34: why did we need to change this?

tests/core/test_state_store.py
```diff
@@ -54,6 +55,7 @@ def test_is_checkpoint_updated():
     assert s2.is_checkpoint_updated() is False
 
     s2.update_state("test-state1")
+    time.sleep(1)
```
Vikas-kum, line 58: why is this needed?

tests/core/test_state_store.py
```diff
@@ -64,6 +66,7 @@ def test_is_checkpoint_updated():
     assert s2.is_checkpoint_updated()
 
     s2.update_state("test-state2")
+    time.sleep(1)
```
Vikas-kum, line 69: why is this needed?

tests/core/test_state_store.py
```diff
@@ -75,10 +78,12 @@ def test_is_checkpoint_updated():
     assert s2.is_checkpoint_updated()
 
     s2.update_state("test-state3")
+    time.sleep(1)
```
Vikas-kum, line 81: same here?

tests/core/test_state_store.py
anirudhacharya: because `update_state` is an async call. 
Even though I am updating `/subdir1/checkpoint_test1.txt` in line 63 after this call, the actual `update_state` happens after `checkpoint_test1.txt` is written. So while we expect line 66 to be `True`, it will return `False` without the 1 second delay.

I tested this out as follows -
```
tests/core/test_state_store.py::test_is_checkpoint_updated 
last_timestamp: 1586394007.8299778 and timestamp of chkpnt file: 1586394007.8291862
```
`last_timestamp` is when `update_state` writes to file to disk(line 57), and `timestamp of chkpnt file` is when the checkpoint file is written to disk(line64).

tests/core/test_state_store.py
anirudhacharya: same as above

tests/core/test_state_store.py
anirudhacharya: same as above.

tests/core/test_state_store.py
anirudhacharya: this is the failed build link - https://tiny.amazon.com/114zqzab7/IsenLink

config/buildspec_zero_code_change.yml
anirudhacharya: we don't need this. it does not cause any failure. But I can remove it. I am not sure how this change got into this PR.

Vikas-kum: please remove if not needed.

tests/core/test_state_store.py
Vikas-kum: why would it fail? you say update_state() , a new json file would be written. 
Since there is no new checkpoint file created, irrespective of whether update_state wrote the file or not youshould get False as result of is_checkpoint_updated.
Am I missing something?

tests/core/test_state_store.py
```diff
@@ -0,0 +1,89 @@
...
+    assert s2.is_checkpoint_updated() is False
+
+    s2.update_state("test-state1")
+    time.sleep(1)
+    # checkpoints_dir still has only metadata.json. It should return false.
+    assert s2.is_checkpoint_updated() is False
+
+    os.mkdir(s2._checkpoint_dir + "/subdir1")
+    with open(s2._checkpoint_dir + "/subdir1/checkpoint_test1.txt", "w") as f:
+        f.write("checkpoint-test-string-1")
```
Vikas-kum, line 64: sync here . f.flush() to make sure that next line is certain to return true.

tests/core/test_state_store.py
Vikas-kum: As I said earlier, I believe you shouldn't need this and below

tests/core/test_state_store.py
```diff
@@ -0,0 +1,89 @@
...
+    # A checkpoint file has been updated. The checkpoint update time is greater than _checkpoint_update_timestamp.
+    # is_checkpoint_updated should return true.
+    assert s2.is_checkpoint_updated()
+
+    s2.update_state("test-state3")
+    time.sleep(1)
+    os.mkdir(s2._checkpoint_dir + "/subdir2")
+    with open(s2._checkpoint_dir + "/subdir2/checkpoint_test2.txt", "w") as f:
+        f.write("checkpoint-test-string-3")
+    # A new checkpoint file has been created. The checkpoint update time is greater than _checkpoint_update_timestamp.
```
Vikas-kum, line 85: f.flush() to make sure that file is written. Also, make sure that flush is right call.

tests/core/test_state_store.py
```diff
@@ -0,0 +1,89 @@
...
+    mock_metadata = [
+        {
+            "training-run": "",
+            "latest-global-step-saved": "",
+            "latest-global-step-seen": "",
+            "latest-mode-step": "",
+        }
+    ]
+    with open(dir_path + "/metadata.json", "w") as f:
+        json.dump(mock_metadata, f)
```
Vikas-kum, line 36: f.flush()

tests/core/test_state_store.py
```diff
@@ -0,0 +1,89 @@
...
+    assert s2.is_checkpoint_updated()
+
+    s2.update_state("test-state2")
+    time.sleep(1)
+    # the state_file has been updated. The lastest checkpoint update time is lesser than _checkpoint_update_timestamp.
+    # is_checkpoint_updated should return false.
+    assert s2.is_checkpoint_updated() is False
+
+    with open(s2._checkpoint_dir + "/subdir1/checkpoint_test1.txt", "a") as f:
+        f.write("checkpoint-test-string-2")
```
Vikas-kum, line 75: flush.

tests/core/test_state_store.py
anirudhacharya: yes you should get False in line 60 and True in line 66. the `time.sleep` is to ensure a True value in line 66.

tests/core/test_state_store.py
anirudhacharya: flush here might not make any difference. 

We could probably remove the sleep if we did a `fileObject.flush` and `os.fsync(f)` here - [state_store.py#L104](https://github.com/awslabs/sagemaker-debugger/blob/master/smdebug/core/state_store.py#L104)

Vikas-kum: yeah, this is what I suggested earlier, please remove the sleep from code and use flush(method which does fsync) . we dont neeed to do fsync in code, but in test

tests/core/test_state_store.py
Vikas-kum: use flush + fsync

Vikas-kum:
https://unix.stackexchange.com/questions/11599/determine-file-system-timestamp-precision#12756
file system time precision is causing this probably. Ok to put sleep before when you create new files ( Write down reason in comments that https://unix.stackexchange.com/questions/11599/determine-file-system-timestamp-precision#12756) 
Put sleep of 2 seconds to be safe.

anirudhacharya:
> https://unix.stackexchange.com/questions/11599/determine-file-system-timestamp-precision#12756
> file system time precision is causing this probably.
> what if same thing happens in prodcution, last_update_timestamp > gmtime and it will say that file was not updated and miss the new checkpoint file if timediff between last_updatetimestamp and gmtime is ~1 sec say

there is a lag between the time reported by python's time `module` and system time reported by `os.path.getmtime()`. Here is a code snippet that demonstrates that 

```python
import time
import os
import json

print(f"t0: {time.time()}")

with open("status.json", 'w') as f:
    json.dump("status", f)

t1 = os.path.getmtime("status.json")
print(f"t1: {t1}")

print(f"t2: {time.time()}")

print(f"t3: {time.time()}")

with open("test.txt", 'w') as f:
    f.write("test")
t4 = os.path.getmtime("test.txt")
print(f"t4: {t4}")
print(f"t5: {time.time()}")
```
Output:
```bash
t0: 1586555897.3936992
t1: 1586555897.388582
t2: 1586555897.3938596
t3: 1586555897.393874
t4: 1586555897.388582
t5: 1586555897.3939729
```

Vikas-kum:
fixed in PR:https://github.com/awslabs/sagemaker-debugger/pull/200

PR #209, opened 2020-04-12
metadata.json may be written again and again

Vikas-kum:
### Description of changes:
During a Sagemaker training job, there was an issue of the `metadata.json` file in the `checkpoints_dir` being re-written every second.

This probably happened because we use `json.dumps` [here](https://github.com/awslabs/sagemaker-debugger/blob/master/smdebug/core/state_store.py#L104) to write to the `states_file` every time the `last_modified_time` of the `checkpoint_dir` is greater than `last_Saved time`( [checked here](https://github.com/awslabs/sagemaker-debugger/blob/master/smdebug/core/state_store.py#L75)). 

`json.dumps` may not guarantee that file was actually written to disk. The file could be in cache and flushed to disk asynchronously a second later. In such a case the `last_modified_time` of the file will always be more than `last_Saved time`. 

To fix it I have filtered out the `states_file` while determining if the `checkpoint_dir` has been updated( [here](https://github.com/awslabs/sagemaker-debugger/blob/master/smdebug/core/state_store.py#L82)).

Filesystems can provide modified time upto certain precision, that is why time.time() approach may not work. 
https://unix.stackexchange.com/questions/11599/determine-file-system-timestamp-precision#12756

So we may not be able to use modified time to determine if files has changed. In this PR we have changed the mechanism to : 
if num_files has changed or file_sizes have changed or if actual file_names have changed. 

It can still happen that file size and file name is same but contents have changed, we have not covered this case because it is not likely. Testing if actual contents have changed can be done though md5sum hashing but it can be costly and un-necessary for this case. 

#### Style and formatting:

### DIFF
--- a/config/buildspec_zero_code_change.yml
+++ b/config/buildspec_zero_code_change.yml
@@ -31,8 +31,8 @@ phases:
 
   build:
     commands:
-      - cd $CODEBUILD_SRC_DIR  && chmod +x config/tests.sh && PYTHONPATH=. && ./config/tests.sh && cd ..
-      - cd $CODEBUILD_SRC_DIR_RULES && chmod +x config/tests.sh && PYTHONPATH=. && ./config/tests.sh  && cd ..
+      - cd $CODEBUILD_SRC_DIR  && chmod +x config/tests.sh && PYTHONPATH=. ./config/tests.sh && cd ..
+      - cd $CODEBUILD_SRC_DIR_RULES && chmod +x config/tests.sh && PYTHONPATH=. ./config/tests.sh  && cd ..
 
   post_build:
     commands:
--- a/config/buildspec_zero_code_change.yml
+++ b/config/buildspec_zero_code_change.yml
@@ -31,8 +31,8 @@ phases:
 
   build:
     commands:
-      - cd $CODEBUILD_SRC_DIR  && chmod +x config/tests.sh && PYTHONPATH=. ./config/tests.sh && cd ..
-      - cd $CODEBUILD_SRC_DIR_RULES && chmod +x config/tests.sh && PYTHONPATH=. ./config/tests.sh  && cd ..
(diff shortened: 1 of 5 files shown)


anirudhacharya:
Thanks @Vikas-kum. 

Just for the record, using `os.path.getmtime(f)` everywhere(including `update_state` function in `state_store.py` would give accurate results as long the timediff between last_update_timestamp and gmtime is `>0.006s`

```python
import time
import os
import json

SLEEP_TIME=0.006

with open("status.json", 'w') as f:
    json.dump("status", f)

t1 = os.path.getmtime("status.json")
print(f"t1: {t1}")

time.sleep(SLEEP_TIME)

with open("test2.txt", 'w') as f:
    f.write("test2")

t2 = os.path.getmtime("test2.txt")
print(f"t2: {t2}")

time.sleep(SLEEP_TIME)

with open("test3.txt", 'w') as f:
    f.write("test3")

t3 = os.path.getmtime("test3.txt")
print(f"t3: {t3}")

time.sleep(SLEEP_TIME)

with open("test4.txt", 'w') as f:
    f.write("test4")
    
t4 = os.path.getmtime("test4.txt")
print(f"t4: {t4}")

time.sleep(SLEEP_TIME)

with open("test5.txt", 'w') as f:
    f.write("test5")
    
t5 = os.path.getmtime("test5.txt")
print(f"t5: {t5}")

time.sleep(SLEEP_TIME)

with open("test6.txt", 'w') as f:
    f.write("test6")
    
t6 = os.path.getmtime("test6.txt")
print(f"t6: {t6}")
```

```bash
t1: 1586732451.4707065
t2: 1586732451.4787064
t3: 1586732451.4827065
t4: 1586732451.4907064
t5: 1586732451.4987063
t6: 1586732451.5027063
```

PR #212, opened 2020-04-13
Paranthesize assert statements (#208)

NihalHarish:
### Description of changes:
- Bug in the way the assert statements were written.
- The parentheses ensures that the if else block is evaluated first.
- Also refactored the code to reuse the `is_tf_2_2` fn

#### Style and formatting:


NihalHarish:
Seems like CI hangs for this PR, but works fine for #208

vandanavk:
did you test this with TF 2.2?

NihalHarish:
> did you test this with TF 2.2?

See #217

Issue #205, opened 2020-04-09
Loss Tensors Are Saved Twice On AWS Pytorch


NihalHarish:
loss functional values are saved twice for each step with AWS Pytorch.

This happens because functional losses are saved by default by the post_hook_for_loss_functional fn in AWS Pytorch.

```
...
    for _ in range(n_steps):
        optimizer.zero_grad()
        outputs = net(inputs)
        loss = F.cross_entropy(outputs, labels)
        hook.record_tensor_value("nll_loss", tensor_value=loss)
        loss.backward()
        optimizer.step()
...
```

The `post_hook_for_loss_functional` is called by `F.cross_entropy(outputs, labels)`

vandanavk:
this happens when, like the test https://github.com/awslabs/sagemaker-debugger/blob/master/tests/pytorch/test_loss.py#L61, user makes use of AWS-pytorch, but also modifies the training script to manually save loss.

PR #213, opened 2020-04-14
Do not save loss and check for loss in Pytorch Data Parallel

NihalHarish:
### Description of changes:
- Dataparallel test checks if smdebug supports pytorch data parallel.
- There is currently a [bug](https://github.com/awslabs/sagemaker-debugger/issues/205) in saving losses 
- Removing logic that saves loss and checks for it in this test, since it is not required to check the sanity of data parallel functionality 

#### Style and formatting:

### DIFF
--- a/tests/pytorch/utils.py
+++ b/tests/pytorch/utils.py
@@ -48,7 +48,6 @@ def train(model, hook, device, optimizer, num_steps=500, set_modes=False):
         optimizer.zero_grad()
         output = model(Variable(data, requires_grad=True))
         loss = F.nll_loss(output, target)
-        hook.record_tensor_value("nll_loss", tensor_value=loss)
         loss.backward()
         optimizer.step()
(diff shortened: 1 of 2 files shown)


NihalHarish:
See #217

tests/pytorch/utils.py
```diff
@@ -48,7 +48,6 @@ def train(model, hook, device, optimizer, num_steps=500, set_modes=False):
         optimizer.zero_grad()
         output = model(Variable(data, requires_grad=True))
         loss = F.nll_loss(output, target)
-        hook.record_tensor_value("nll_loss", tensor_value=loss)
```
vandanavk, line 51: if this line is removed, will it still save loss for vanilla PT ?

NihalHarish: Reverting this.

PR #215, opened 2020-04-14
Refactor Zero Code Change Test Directory

NihalHarish:
### Description of changes:
- Moved file constants to the `constants.py` file
- Moved the launch fn to the `utils.py` file
- Also changed horovodrun to mpirun becuase of possible bug in DLC containers. See [internal-ticket](https://t.corp.amazon.com/DLAMI-771/overview)

#### Style and formatting:

I have run `pre-commit install` to ensure that auto-formatting happens with every commit.

#### Issue number, if available

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #156, opened 2020-01-23
Tests to Check Pytorch Hook Sanity With Horovod Training and Zero Code Change

NihalHarish:
### Description of changes:
- Pytorch Horovod script that runs training 
- Tests that check the sanity of the metrics collected by smdebug
- Refactored existing TF Horovod tests to re-use code

**Closed:  See** #216 

#### Style and formatting:

I have run `pre-commit install` to ensure that auto-formatting happens with every commit.

#### Issue number, if available

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #216, opened 2020-04-14
Pytorch Horovod Zero Code Change Test

NihalHarish:
### Description of changes:
- Pytorch Horovod script that runs distributed training
- Note: This PR is rebased off of #215 
- This incorporates the reviews received on  #156, 

#### Style and formatting:


examples/pytorch/zero_code_change_examples/horovod_mnist.py
```diff
@@ -0,0 +1,198 @@
...
+
+    # Horovod: broadcast parameters & optimizer state.
+    hvd.broadcast_parameters(model.state_dict(), root_rank=0)
+    hvd.broadcast_optimizer_state(optimizer, root_rank=0)
+
+    # Horovod: (optional) compression algorithm.
+    compression = hvd.Compression.none
+
+    # Horovod: wrap optimizer with DistributedOptimizer.
+    optimizer = hvd.DistributedOptimizer(
```
Vikas-kum, line 192: optimizer is not passed to train function ? 
train_function uses this optimizer in L53 . 
Wondering how is this working? Is this global static object?

NihalHarish: Interesting question.

I double checked with the [example](https://github.com/horovod/horovod/blob/master/examples/mxnet_mnist.py) and they seem to be doing the same thing.

I also defined an optimizer outside of the train function with a vanilla pytorch example and it failed with `NameError: name 'optimizer' is not defined` as expected.

I think in this case, hvd is taking care of the distribution and availability of the optimizer, since we have run `hvd.init()` before the definition of the optimizer and we also wrap the optimizer with `hvd.DistributedOptimizer`

Vikas-kum: Then all examples are wrong, probably?
"I think in this case, hvd is taking care of the distribution and availability of the optimizer, since we have run hvd.init() before the definition of the optimizer and we also wrap the optimizer with hvd.DistributedOptimizer"

I didn't understand this. 
There is local variable optimizer, please figure out the right way to do it. Or say optimizer is global ?
Who is creating that global object ? 
Does it still work if you change optimizer to optimizer1 in all your source code?

NihalHarish: Looks like its a feature of python. 

Links: 
[DOCS](https://docs.python.org/3/tutorial/classes.html#python-scopes-and-namespaces)
[S/O answer](https://stackoverflow.com/questions/4775579/main-and-scoping-in-python).


I tested this snippet in Python 3


```
def test_random_var():
    assert __name__ == "__main__"
    assert random_var == 5008
    assert 'random_var' in globals()
    print("Pass")

if __name__ ==  "__main__":
    random_var = 5008
    test_random_var()
```

It succeeds when we run `python test.py` and fails for `pytest test.py`.
Because `__name__ == "test"` in the latter.

PR #221, opened 2020-04-20
Modifying the assert to work on both vanilla and AWS PT

vandanavk:
### Description of changes:
Vanilla PT will save loss only once when hook.record_tensor_value() is called.
While for AWS-PT, this test saves loss twice as described in this issue https://github.com/awslabs/sagemaker-debugger/issues/205.

The flavor of PT cannot be differentiated in the code, hence modifying the assert in the test to work with both flavors.

#### Style and formatting:

### DIFF
--- a/tests/pytorch/test_loss.py
+++ b/tests/pytorch/test_loss.py
@@ -78,8 +78,8 @@ def test_register_loss_functional(out_dir):
     loss_tensor = trial.tensor("nll_loss_output_0")
 
     # Capture ['nll_loss_output_0']
-    assert len(trial.tensor_names()) == 2
-    assert len(loss_coll.tensor_names) == 2
+    assert len(trial.tensor_names()) >= 1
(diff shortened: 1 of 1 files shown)


tests/pytorch/test_loss.py
```diff
@@ -78,8 +78,8 @@ def test_register_loss_functional(out_dir):
     loss_tensor = trial.tensor("nll_loss_output_0")
 
     # Capture ['nll_loss_output_0']
-    assert len(trial.tensor_names()) == 2
-    assert len(loss_coll.tensor_names) == 2
+    assert len(trial.tensor_names()) >= 1
+    assert len(loss_coll.tensor_names) >= 1
```
NihalHarish, line 82: Can you add a comment here with a reference to the issue?

vandanavk: done in the latest commit

PR #225, opened 2020-04-22
Log Unsupported TF Variables

NihalHarish:
### Description of changes:
- Add a string output to the NotImplemented Error so that it helps us identify the exact type of variable that causes failures.

#### Style and formatting:

### DIFF
--- a/smdebug/tensorflow/tensor_ref.py
+++ b/smdebug/tensorflow/tensor_ref.py
@@ -19,7 +19,7 @@ def get_tf_names(arg):
         tf_names = [arg.name]
     elif isinstance(arg, tf.Tensor):
         tf_names = [arg.name]
-    elif isinstance(arg, values.DistributedVariable):
+    elif isinstance(arg, values.DistributedValues):
         tf_names = [v.name for v in arg._values]
     else:
         raise NotImplementedError(
--- a/smdebug/tensorflow/tensor_ref.py
+++ b/smdebug/tensorflow/tensor_ref.py
@@ -22,7 +22,7 @@ def get_tf_names(arg):
     elif isinstance(arg, values.DistributedVariable):
         tf_names = [v.name for v in arg._values]
     else:
-        raise NotImplementedError
+        raise NotImplementedError(f"Smdebug currenty not support:{arg} which of type:{type(arg)}")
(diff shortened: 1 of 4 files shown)


smdebug/tensorflow/tensor_ref.py
```diff
@@ -22,7 +22,9 @@ def get_tf_names(arg):
     elif isinstance(arg, values.DistributedVariable):
         tf_names = [v.name for v in arg._values]
     else:
-        raise NotImplementedError
+        raise NotImplementedError(
```
Vikas-kum, line 25: should we just support distributed values I don't see things getting worse than what it is today.
a  proposal is: 
else:
  #log
  if instance of DistributedValue : 
      tf_names = [v.name for v in arg._values] 
 else:
   Not implemented error

NihalHarish: Done.

smdebug/tensorflow/tensor_ref.py
```diff
@@ -19,10 +19,12 @@ def get_tf_names(arg):
         tf_names = [arg.name]
     elif isinstance(arg, tf.Tensor):
         tf_names = [arg.name]
-    elif isinstance(arg, values.DistributedVariable):
+    elif isinstance(arg, values.DistributedValues):
```
vandanavk, line 22: make sure you replace all occurrences of DistributedVariable. also, make sure it doesn't break the training script reported on a recent internal ticket

NihalHarish: Done. Thanks for this

NihalHarish:
> we should probably have a test or a sample training script for this case.

Don't the current tests and examples execute this same code path.

vandanavk:
> > we should probably have a test or a sample training script for this case.
> 
> Don't the current tests and examples execute this same code path?

Doesn't look like the current tests test all child classes of DistributedValues.

Vikas-kum:
> > > we should probably have a test or a sample training script for this case.
> > 
> > 
> > Don't the current tests and examples execute this same code path?
> 
> Doesn't look like the current tests test all child classes of DistributedValues.

Yeah, seee if there i

PR #228, opened 2020-04-29
Add Assert Statements that track the default tensors saved in ZCC

NihalHarish:
### Description of changes:
- Added assert statements to track default tensors saved in ZCC
- Assert statements have intentionally not been added to the grad tape test because no tensors are saved in ZCC and creating a trial will create a an infinite wait.

#### Style and formatting:

### DIFF
--- a/tests/zero_code_change/tensorflow2_integration_tests.py
+++ b/tests/zero_code_change/tensorflow2_integration_tests.py
@@ -22,6 +22,7 @@ import tensorflow.compat.v2 as tf
 
 # First Party
 import smdebug.tensorflow as smd
+from smdebug.core.collection import CollectionKeys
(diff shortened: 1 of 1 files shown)


tests/zero_code_change/utils.py
```diff
@@ -54,3 +54,10 @@ def build_json(
     with open(path.absolute(), "w") as outfile:
         json.dump(config_dict, outfile)
     return path.absolute()
+
+
+def launch_job(script_file_path, script_args, config_file_path):
```
Vikas-kum, line 59: where is this used?

NihalHarish: This is not used anywhere.
I have squashed and rebased the branch.
Apologies for the wrong diff.

PR #237, opened 2020-05-06
Save Opt Var in TRAIN mode only

NihalHarish:
### Description of changes:
- `Bug Description`: Optimizer variables are only recorded during TRAIN mode.
However, while recording tensors, when `set_optimizer_variables` was invoked, `self.mode` inconsistently TRAIN or GLOBAL across Vanilla and AWS TF.
- This PR explicitly saves optimizer variables with `mode=ModeKeys.TRAIN`

#### Style and formatting:


vandanavk:
Can we have a test to catch this scenario?

NihalHarish:
> Can we have a test to catch this scenario?

This error was caught only because tests were failing on sagemaker-debugger-rules, which executes tests with Vanilla TF 1.15

PR #242, opened 2020-05-08
Update GradientTape Integration Tests 

NihalHarish:
### Description of changes:
- Update assert statements to check for ZCC functionality with AWS ZCC

#### Style and formatting:


tests/zero_code_change/test_tensorflow2_gradtape_integration.py
```diff
@@ -97,7 +97,13 @@ def helper_test_keras_v2_gradienttape(script_mode: bool = False, json_file_conte
                 print(log)
                 train_acc_metric.reset_states()
             hook = smd.get_hook()
-            assert not hook
+            assert hook
+            hook.close()
+            # Check that hook created and tensors saved
+            trial = smd.create_trial(path=sim.out_dir)
+            assert len(trial.steps()) > 0, "Nothing saved at any step."
```
vandanavk, line 104: these might fail on older DLC which dont support ZCC. can you add a check to do this check only if hook is valid?

tests/zero_code_change/test_tensorflow2_gradtape_integration.py
```diff
@@ -97,7 +98,15 @@ def helper_test_keras_v2_gradienttape(script_mode: bool = False, json_file_conte
                 print(log)
                 train_acc_metric.reset_states()
             hook = smd.get_hook()
-            assert not hook
+            if not is_tf_2_2():
+                assert not hook  # only supported on TF 2.2 and greater
```
vandanavk, line 102: return after assert?

Issue #255, opened 2020-05-20
keras TF 2.2 mileading error mesg


Vikas-kum:
Got this error mesg - 
 Disabling SMDebug as it does not support eager modefor TF versions 1.x

>>> import tensorflow as tf
>>> tf.__version__
‘2.2.0' 

I am using run_eagerly=True in model compile. 

model.compile(
      loss='sparse_categorical_crossentropy',
      optimizer=tf.keras.optimizers.Adam(0.001),
      metrics=['accuracy’], run_eagerly=True
  )

PR #256, opened 2020-05-21
Fix Incorrect Log Statement

NihalHarish:
### Description of changes:
- Log was being printed for TF 2.x versions although it was only supposed to appear for 1.x versions.
- Adding parentheses to make the boolean logic for explicit

#### Style and formatting:

#255 


vandanavk:
LGTM, can you add a test in tensorflow2/test_keras.py ?

one test that passes run_eagerly=True to helper_keras_fit() that sets this in model.compile

PR #257, opened 2020-05-22
Incorrect number of tensors saved with MirroredStrategy

NihalHarish:
### Description of changes:
- Bug with `_prepare_non_layer_tensors`. The outerloop was iterating over all the collections instead of just gradients and optimizer variables and adding the wrong number of tensors into custom collections.
- With `mirroredstrategy` and a saveconfig with only optimizer_variable collections; writers are not initialized, this has been corrected with an explicit ` self._initialize_writers(only_initialize_if_missing=True)` call in `_save_optimizer_variables`
- `test_include_regex` now tests with include_workers= `one` and `all`
- New test `test_include_regex_opt_var` that attempts to save only optimizer variables with a custom collection 

#### Style and formatting:

### DIFF
--- a/tests/tensorflow2/test_keras_mirrored.py
+++ b/tests/tensorflow2/test_keras_mirrored.py
@@ -390,12 +390,13 @@ def test_training_end(out_dir, tf_eager_mode):
 
 
 @pytest.mark.slow
-def test_include_regex(out_dir, tf_eager_mode):
+@pytest.mark.parametrize("workers", ["one", "all"])
+def test_include_regex(out_dir, tf_eager_mode, workers):
     hook = KerasHook(
         out_dir=out_dir,
         save_config=SaveConfig(save_interval=9),
         include_collections=["custom_coll"],
-        include_workers="all",
+        include_workers=workers,
     )
     hook.get_collection("custom_coll").include("dense")
     strategy = train_model(out_dir, hook=hook, steps=["train"], eager=tf_eager_mode)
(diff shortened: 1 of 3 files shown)


tests/tensorflow2/test_keras_mirrored.py
```diff
@@ -390,12 +390,13 @@ def test_training_end(out_dir, tf_eager_mode):
...
 @pytest.mark.slow
-def test_include_regex(out_dir, tf_eager_mode):
+@pytest.mark.parametrize("workers", ["one", "all"])
+def test_include_regex(out_dir, tf_eager_mode, workers):
     hook = KerasHook(
         out_dir=out_dir,
         save_config=SaveConfig(save_interval=9),
         include_collections=["custom_coll"],
-        include_workers="all",
+        include_workers=workers,
```
vandanavk, line 399: do the number of tensors remain 4 when workers=one and workers=all? probably should vary with strategy.num_replicas_in_sync in the case of workers=all

vandanavk: can you check if we have an issue in general with save_all_workers in TF eager mode? maybe parameterizing all the tests to test workers="one" and workers="all" would help to check quickly

NihalHarish: > do the number of tensors remain 4 when workers=one and workers=all? probably should vary with strategy.num_replicas_in_sync in the case of workers=all

The number of tensors will remain the same.
I have added an assert to count the number of workers for each tensor

NihalHarish: > can you check if we have an issue in general with save_all_workers in TF eager mode? maybe parameterizing all the tests to test workers="one" and workers="all" would help to check quickly

Are we worried at all about slowing down the test suite by doing this though?

vandanavk: maybe add this parameterization only for test_save_all?

NihalHarish: There seems to be three tests that seems to be doing the same thing:

`test_save_all_workers`, `test_save_one_worker` and `test_save_all_workers`.

Maybe we should coalesce these three by just parameterizing `test_save_all_workers`?

vandanavk: sounds good

tests/tensorflow2/test_keras_mirrored.py
```diff
@@ -260,6 +262,10 @@ def test_save_all(out_dir, tf_eager_mode):
         )
         # weights, grads, optimizer_variables, metrics, losses, outputs
     assert len(tr.steps()) == 3
+    for tname in tr.tensor_names():
+        assert len(tr.tensor(tname).workers(0)) == (
+            1 if workers == "one" else strategy.num_replicas_in_sync
+        )
```
vandanavk, line 268: now that you made this change,  can you merge the tests test_save_one_worker and test_save_all_workers here combining all the relevant assert statements?

NihalHarish: done

PR #259, opened 2020-05-26
Save Scalars With Mirrored Strategy

NihalHarish:
### Description of changes:
- Saved scalars do not have tensor_refs
- When users tried to save scalars while training with Mirrored strategy they ran into errors.
- This code change adds supports for scalars

#### Style and formatting:


tests/tensorflow2/test_keras_mirrored.py
```diff
@@ -124,6 +124,7 @@ def scale(image, label):
         steps = ["train"]
     for step in steps:
         if step == "train":
+            hook.save_scalar("epoch", 1, sm_metric=True)
```
vandanavk, line 127: can you verify that the scalar has been written using tests.core.utils.verify_files? at least in one test?

vandanavk: also add an example to examples/tensorflow2 covering this case?

smdebug/core/hook.py
```diff
@@ -653,7 +653,13 @@ def _write_scalars(self):
                     )
             if write_event:
                 if self.writer is None:
-                    self._initialize_writers()
+                    try:
+                        self._initialize_writers()
+                    except NotImplementedError:
+                        # This can happen when the user explicitly calls save_scalars
```
Vikas-kum, line 659: not sure if I understand this comment?

NihalHarish: User Script: 
```
hook = (
...
)
strategy = UndefinedTFStrategy() # strategy not currently supported by smdebug
train() # hook does not save any data. Prints a log msg and turns off
hook.save_scalar("foobar") # Error due to error thrown by line 656
```

tests/tensorflow2/test_keras_mirrored.py
NihalHarish: > can you verify that the scalar has been written using tests.core.utils.verify_files? at least in one test?

Done

smdebug/core/hook.py
vandanavk: Is "training method not supported" applicable only to TF or other frameworks too? 

when a strategy is not supported, we set self._hook_supported to False. in general, what happens when the user calls, say, `hook.wrap_optimizer` or `hook.record_tensor_value` when hook is not supported?

vandanavk: also, why is this only relevant here and not elsewhere where initialize_writers is called?

can you move this NotImplementedError to the definition of initialize_writers() if it is applicable in other scenarios too?

NihalHarish: > when a strategy is not supported, we set self._hoo

If `hook._is_not_supported() is True` then we do not save any data.
This check is currently done at the start of all the callbacks.

Data saved by both `hook.wrap_optimizer` and `hook.record_tensor_value` are saved during these callbacks.

save_scalars however writes data outside the callback codepath and hence requires additional checks to ensure that we do not save any data.

smdebug/core/hook.py
```diff
@@ -652,8 +654,17 @@ def _write_scalars(self):
...
-                if self.writer is None:
+                writer_not_initialized = False
+                if isinstance(self, BaseHook):
+                    writers = self._get_writers(None, None)
+                    if not any(writers):
+                        writer_not_initialized = True
+                elif self.writer is None:
+                    writer_not_initialized = True
+
+                if writer_not_initialized:
```
vandanavk, line 665: instead of these if checks, would it be possible to modify core/hook.py:_initialize_writers to   _initialize_writers(self, only_initialize_if_missing=False) ?

that way you can just call    self._initialize_writers(self, only_initialize_if_missing=True) here?

NihalHarish: You're right. This is much cleaner.

smdebug/core/hook.py
```diff
@@ -652,8 +658,10 @@ def _write_scalars(self):
                         scalar_name, scalar_val, self.step, timestamp=timestamp
                     )
             if write_event:
-                if self.writer is None:
-                    self._initialize_writers()
+                if self._is_not_supported():
```
vandanavk, line 661: if hook is not supported, don't add it to ScalarCache? in save_scalar()

NihalHarish: Done.

smdebug/core/hook.py
```diff
@@ -294,6 +294,10 @@ def _get_worker_name(self):
     def _get_num_workers(self):
         pass
 
+    @abstractmethod
+    def _is_not_supported(self):
```
Vikas-kum, line 298: I dont see the impl

smdebug/core/hook.py
```diff
@@ -652,8 +661,7 @@ def _write_scalars(self):
                         scalar_name, scalar_val, self.step, timestamp=timestamp
                     )
             if write_event:
-                if self.writer is None:
-                    self._initialize_writers()
+                self._initialize_writers(only_initialize_if_missing=True)
```
Vikas-kum, line 661: how is this different that what was there earlier.
it will be no op if self.writer is not None which is same as earlier behavior.

smdebug/core/hook.py
```diff
@@ -668,6 +673,9 @@ def save_scalar(self, name, value, sm_metric=False, timestamp: float = None):
         :param sm_metric: True/False. If set to True, the scalar value will be written to
         SageMaker
         """
+        if self._is_not_supported():
+            # Do not log scalars if smdebug hook is not supported
```
Vikas-kum, line 677: smdebug hook is not supported means? What is the case?

smdebug/core/hook.py
NihalHarish: This is an abstractmethod.
Impl here:
https://github.com/awslabs/sagemaker-debugger/blob/01c96b2ce487ecdea336387263f64a59d38f22d7/smdebug/tensorflow/session.py#L237
https://github.com/awslabs/sagemaker-debugger/blob/f1498e42f72bc4381960b350ddb0b5b46b397a5d/smdebug/tensorflow/keras.py#L61

smdebug/core/hook.py
NihalHarish: `self.writer` is always `None` for mirrored strategy.

So, in the earlier case, `self._initialize_writers()` was always creating new writers and deleting the existing writers with buffered data.

`self._initialize_writers(only_initialize_if_missing=True)` calls https://github.com/awslabs/sagemaker-debugger/blob/e99b7054f4be814c751890344969a825f257f9cb/smdebug/tensorflow/base_hook.py#L305

which handles the mirrored strategy case

smdebug/core/hook.py
NihalHarish: If smdebug hook is not supported by the training strategy.

See: https://github.com/awslabs/sagemaker-debugger/blob/e6359444d28ae7de260074ef4d5962ac83fa649a/tests/tensorflow2/test_keras_mirrored.py#L417

smdebug/core/hook.py
Vikas-kum: say -  if self.writer is None for mirrored strategy, 
This line         if only_initialize_if_missing and self.writer: ==> will evaluate to false. and then self.writer would be inited in L449, 
The same thing will happen if this code change is not there. Am i missing anything here.

NihalHarish: Yes, that is correct.
But before my code change, we had two 
`_initialize_writers` fns. One that accepted only_initialize_if_missing and one that didn't.
I have simply just unified the two of them.

NihalHarish: Here is the snippet of my change to the `_initialize_writers` fn:

```
    def _initialize_writers(self, only_initialize_if_missing=False) -> None:
        if only_initialize_if_missing and self.writer:
            return
```

NihalHarish: Link to the impl in base_hook.

https://github.com/awslabs/sagemaker-debugger/blob/e6359444d28ae7de260074ef4d5962ac83fa649a/smdebug/tensorflow/base_hook.py#L305

Vikas-kum: can you write a comment on core hook initialize writers that this method can be overridden. For tensorflow, the method is overridden in baseHook class.

PR #263, opened 2020-05-28
Don't export to collections for all workers with unsupported distrib training

vandanavk:
### Description of changes:
When users make use of unsupported distributed training such as python multiprocessing, smdebug writes tensors captures by the first process. Looks like all the processes were trying to write collections.json. Fixing that in this PR.

Also, removing the warnings that were printed frequently

#### Style and formatting:

### DIFF
--- a/smdebug/core/hook.py
+++ b/smdebug/core/hook.py
@@ -436,6 +436,12 @@ class BaseHook:
                     self.first_process = True
                     self.logger.info(f"Hook is writing from the hook with pid: {os.getpid()}\n")
                 else:
+                    if self.first_process is None:
+                        self.logger.warn(
+                            f"Unsupported Distributed Training Strategy Detected. \
+                            Sagemaker-Debugger will only write from one process. \
+                            The process with pid: {os.getpid()} will not be writing any data. \n"
+                        )
                     self.first_process = False
                     return
 
@@ -542,6 +548,11 @@ class BaseHook:
     def export_collections(self):
         num_workers = self._get_num_workers()
         if num_workers == 1 and self.first_process is False:
+            self.logger.warn(
+                f"Unsupported Distributed Training Strategy Detected. \
+                Sagemaker-Debugger will only write from one process. \
+                The process with pid: {os.getpid()} will not be writing any data. \n"
+            )
             return
         if self.save_all_workers is False:
             if self.chief_worker != self.worker:
--- a/smdebug/core/hook.py
+++ b/smdebug/core/hook.py
@@ -437,11 +437,6 @@ class BaseHook:
                     self.logger.info(f"Hook is writing from the hook with pid: {os.getpid()}\n")
                 else:
                     self.first_process = False
-                    self.logger.warn(
-                        f"Unsupported Distributed Training Strategy Detected.\n\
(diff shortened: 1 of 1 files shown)


smdebug/core/hook.py
```diff
@@ -437,11 +437,6 @@ def _initialize_writers(self, only_initialize_if_missing=False) -> None:
                     self.logger.info(f"Hook is writing from the hook with pid: {os.getpid()}\n")
                 else:
                     self.first_process = False
-                    self.logger.warn(
```
Vikas-kum, line 440: move this log to 439 and then we will be logging only once. 
if self.first_process is None:
  log
self.first_process= False

smdebug/core/hook.py
```diff
@@ -546,6 +541,8 @@ def set_mode(self, mode):
 
     def export_collections(self):
         num_workers = self._get_num_workers()
+        if num_workers == 1 and self.first_process is False:
+            return
```
Vikas-kum, line 556: similar log statement here.

smdebug/core/hook.py
```diff
@@ -546,6 +541,8 @@ def set_mode(self, mode):
 
     def export_collections(self):
         num_workers = self._get_num_workers()
+        if num_workers == 1 and self.first_process is False:
```
NihalHarish, line 544: `self._get_num_workers()` returns one of the three values:
```
1. torch.distributed.get_world_size()
2. hvd.size()
3. 1
```
When is the value ≠ 1

smdebug/core/hook.py
```diff
@@ -437,11 +437,6 @@ def _initialize_writers(self, only_initialize_if_missing=False) -> None:
                     self.logger.info(f"Hook is writing from the hook with pid: {os.getpid()}\n")
                 else:
                     self.first_process = False
-                    self.logger.warn(
-                        f"Unsupported Distributed Training Strategy Detected.\n\
-                        Sagemaker-Debugger will only write from one process.\n\
-                        The process with pid: {os.getpid()} will not be writing any data. \n"
-                    )
```
NihalHarish, line 444: We should probably move this log to some other location where it is called fewer times instead of completely removing it?

smdebug/core/hook.py
vandanavk: options 1 and 2 may return more than 1?

smdebug/core/hook.py
vandanavk: fixed this in the commit i just pushed

smdebug/core/hook.py
NihalHarish: Doesn't this then this mean that the distribution strategy is supported?

Vikas-kum: self.first_process is set only when num_workers =1 , so just checking self.first_process = False should suffice. 
https://github.com/awslabs/sagemaker-debugger/commit/62a4e2ab8ea315e9ad575a06f6493bb9c2620b18

smdebug/core/hook.py
vandanavk: added in the latest commit

smdebug/core/hook.py
vandanavk: added in the latest commit

PR #267, opened 2020-06-03
Get Model Inputs and Outputs: Labels and Y Predicate

NihalHarish:
### Description of changes:
- This PR runs only with a modified AWS TF, which supplies `x`, `y` and `y_pred` through logs.
- A new function `_save_model_outputs` that saves outputs to `OUTPUTS` collection 
- Updates to the assert statements to check for the additional tensors we save with this change.

### Performance Testing:
Machine: p3.16x
epochs=5000, steps_per_epoch=10
save_config = SaveConfig(save_interval=500)

Before:
Outputs are not saved:
Uses TF without change
```
real	20m32.546s
user	409m1.362s
sys	456m32.768s
```

With PR: 
Outputs saved:
Uses TF with change
```
real	21m0.540s
user	414m49.772s
sys	461m42.581s
```
#### Style and formatting:


smdebug/tensorflow/base_hook.py
```diff
@@ -437,7 +437,7 @@ def _write_for_tensor(self, tensor_name, tensor_value, save_collections, tensor_
 
         # this tensor_name is tf tensor name, need to convert to export_name
         tensor_ref = self._get_tensor_ref(tensor_name, save_collections=save_collections)
-        if tensor_ref:
+        if tensor_ref is not None:
```
NihalHarish, line 440: Unrelated to PR.
This if check sometimes leads to an annoying error during development.
See: https://stackoverflow.com/questions/7816363/if-a-vs-if-a-is-not-none for more details

Vikas-kum:
Please fix the build failure.

smdebug/tensorflow/keras.py
```diff
@@ -366,6 +366,21 @@ def _add_metric(self, metric_name, metric_value: tf.Tensor = None):
...
+        if self._is_collection_being_saved_for_step(CollectionKeys.OUTPUTS):
+            export_names = {"y_pred": "train_output/y_pred", "y": "train_output/y"}
+            self._initialize_writers(only_initialize_if_missing=True)
+            output_collection = self.collection_manager.get(CollectionKeys.OUTPUTS)
+            for key in logs:
+                if key in ["y", "y_pred"]:
+                    tensor_ref = TensorRef.from_non_graph_var(export_names[key])
+                    output_collection.set_tensor_ref(tensor_ref)
+                    self.tensor_to_collections[export_names[key]] = {output_collection}
+                    self._save_for_tensor(export_names[key], logs[key], check_before_write=False)
```
Vikas-kum, line 382: can we create a save_tensor function. 
save_tensor(name, value, collection=Collection.DEFAULT_tensors)

smdebug/tensorflow/keras.py
```diff
@@ -375,7 +390,7 @@ def _save_metrics(self, batch, logs, force_save=False):
             self._initialize_writers(only_initialize_if_missing=True)
             logs["batch"] = batch
             for key in logs:
-                if key in ["loss", "val_loss", "outputs"]:
+                if key in ["loss", "val_loss", "outputs", "y", "y_pred"]:
```
Vikas-kum, line 393: y ypred is special name used at 374 378 and here. Can we make it as named constant?

smdebug/tensorflow/keras.py
```diff
@@ -479,6 +495,7 @@ def on_epoch_end(self, batch, logs=None):
         if self._is_not_supported():
             return
         self._save_metrics(batch=batch, logs=logs, force_save=True)
+        self._save_model_outputs(logs=logs)
```
Vikas-kum, line 498: if save tensor post step has this logic, why do we need to do it here?

Vikas-kum:
Can you comment what is the value and y and y_pred looks like ?

NihalHarish:
> Please fix the build failure.

The build failure is due to the fact that the pipelines currently do not use the custom AWS TF version

smdebug/tensorflow/keras.py
NihalHarish: We do not need it here, since we currently save values produced during the train step.

smdebug/tensorflow/keras.py
NihalHarish: I have added constants

smdebug/tensorflow/keras.py
NihalHarish: An API for the customers to use?
We can.

NihalHarish:
> Can you add to testing section indescription what is and y and y_pred value looks like ?
> trial.tensor("y").steps()
> trial.tensor("y_pred").value(10)

I have updated the test with additional asserts:

```
        for tname in trial.tensor_names(collection=CollectionKeys.OUTPUTS):
            output = trial.tensor(tname)
            assert tname in [ModelOutput.Y_PRED, ModelOutput.Y]
            assert output.value(0) is not None
            assert output.steps() == trial.steps(mode=ModeKeys.TRAIN)
        # Check the shape of output tensors
        assert trial.tensor(ModelOutput.Y).value(0).shape[1] == 1  # label
        assert (
            trial.tensor(ModelOutput.Y_PRED).value(0).shape[1] == 10
        )  # Output probability for each class
```

smdebug/tensorflow/keras.py
Vikas-kum: Why not other stages (EVAL & predict) ?

Vikas-kum:
> > Please fix the build failure.
> 
> The build failure is due to the fact that the pipelines currently do not use the custom AWS TF version

So, how are we taking care of this? We can't have the test failing. You need to have separate test and enable those only when we have AWS TF changes.

tests/tensorflow2/test_keras_mirrored.py
```diff
@@ -256,8 +256,10 @@ def test_save_all(out_dir, tf_eager_mode, workers):
     tr = create_trial_fast_refresh(out_dir)
     print(tr.tensor_names())
     if tf_eager_mode:
-        assert len(tr.tensor_names()) == (6 + 2 + 1 + 5 + 1 if is_tf_2_2() else 6 + 3 + 1 + 5 + 1)
-        # weights, metrics, losses, optimizer variables, scalar
+        assert len(tr.tensor_names()) == (
+            6 + 2 + 1 + 5 + 1 + 2 if is_tf_2_2() else 6 + 3 + 1 + 5 + 1 + 2
```
Vikas-kum, line 260: this type of test 6+ 2 + 1 has no value, how do we know what is six tensors what is 2 tensors ? It seems cryptic and difficult to maintain.
We should have asserted tname.count == 6
and blah blah, rather than adding 6 + 2 + 1 ...

tests/tensorflow2/test_keras.py
```diff
@@ -396,7 +397,8 @@ def test_gradtape_persistent(out_dir, saveall):
 @pytest.mark.slow
 @pytest.mark.parametrize("saveall", [True, False])
 def test_keras_fit(out_dir, tf_eager_mode, saveall):
-    hook = smd.KerasHook(out_dir=out_dir, save_all=saveall)
+    save_config = SaveConfig(save_interval=1) if saveall else None
```
vandanavk, line 400: might be overkill for CI?

vandanavk:
> > Please fix the build failure.
> 
> The build failure is due to the fact that the pipelines currently do not use the custom AWS TF version

so the tests would fail on vanilla TF too?

tests/tensorflow2/test_keras.py
NihalHarish: We run only 10 steps 

> > > Please fix the build failure.
> > 
> > 
> > The build failure is due to the fact that the pipelines currently do not use the custom AWS TF version
> 
> so the tests would fail on vanilla TF too?

Yes, this feature will only be supported on AWS TF

vandanavk: so the vanilla container codebuild will fail. can you add checks to ensure that it works for both versions of TF ?

smdebug/core/reductions.py
```diff
@@ -29,14 +29,15 @@ def get_basic_numpy_reduction(reduction_name, numpy_data):
...
 
-        if abs:
-            rv = np.linalg.norm(np.absolute(numpy_data), ord=ord)
-        else:
-            rv = np.linalg.norm(numpy_data, ord=ord)
+        if np.isscalar(numpy_data):
+            # np.linalg.norm expects array-like inputs
+            # but numpy_data can sometimes be a scalar value
+            numpy_data = [numpy_data]
+        rv = np.linalg.norm(numpy_data, ord=order)
```
vandanavk, line 40: now that your PR is merged, you dont need this here right?

smdebug/tensorflow/keras.py
```diff
@@ -366,6 +374,42 @@ def _add_metric(self, metric_name, metric_value: tf.Tensor = None):
...
+
+        for key in logs:
+            if key in [ModelOutput.Y, ModelOutput.Y_PRED]:
+                collections_to_save = self._get_collections_to_save_for_step()
+                output_collection = self.get_collection(CollectionKeys.OUTPUTS)
+                if output_collection in collections_to_save:
+                    collections_to_write = {output_collection}
+                    for collection in collections_to_save:
+                        if match_inc(export_names[key], collection.include_regex):
+                            collections_to_write.add(collection)
```
vandanavk, line 396: isn't this done in save_for_Tensor already?

```
save_collections = self._get_collections_with_tensor(tensor_name)
        save_collections_for_tensor = save_collections.intersection(
            self._get_collections_to_save_for_step()
        )
```

smdebug/tensorflow/keras.py
```diff
@@ -95,6 +97,11 @@ def _is_not_supported(self):
                 self._hook_supported = False
         return not self._hook_supported
 
+    def should_save_global_step_for_mode(self, mode: str):
```
vandanavk, line 100: where is this used?

smdebug/core/modes.py
```diff
@@ -14,3 +14,16 @@ class ModeKeys(Enum):
 ALLOWED_MODE_NAMES = [x.name for x in ALLOWED_MODES]
 MODE_STEP_PLUGIN_NAME = "mode_step"
 MODE_PLUGIN_NAME = "mode"
+
+
+def str_to_mode_keys(s):
```
vandanavk, line 19: this is being used only in should_save_global_step_for_mode, which doesn't have a caller. is this required?

NihalHarish: `should_save_global_step_for_mode` is called in aws-tf. See https://tiny.amazon.com/ljsp41dd/IsenLink for details.

smdebug/core/reductions.py
NihalHarish: Weird. This diff is wrong. The RHS doesn't reflect the current state of master.

PR #268, opened 2020-06-03
TF 2.x: Support for keras to estimator

vandanavk:
### Description of changes:
This PR makes a few fixes to enable keras to estimator in TF 2

#### Style and formatting:

### DIFF
--- a/docs/tensorflow.md
+++ b/docs/tensorflow.md
@@ -18,9 +18,14 @@
 - This library itself supports the following versions when you use our API which requires a few minimal changes to your training script: TensorFlow 1.14, 1.15, 2.0.1, 2.1.0. Keras 2.3.
 
 ### Interfaces
-- [Estimator](https://www.tensorflow.org/versions/r1.15/api_docs/python/tf/estimator)
-- [tf.keras](https://www.tensorflow.org/versions/r1.15/api_docs/python/tf/keras)
-- [MonitoredSession](https://www.tensorflow.org/versions/r1.15/api_docs/python/tf/train/MonitoredSession?hl=en)
+- TF 1.x:
+    - [Estimator](https://www.tensorflow.org/versions/r1.15/api_docs/python/tf/estimator)
+    - [tf.keras](https://www.tensorflow.org/versions/r1.15/api_docs/python/tf/keras)
+    - [MonitoredSession](https://www.tensorflow.org/versions/r1.15/api_docs/python/tf/train/MonitoredSession?hl=en)
+- TF 2.x:
+    - [Estimator](https://www.tensorflow.org/versions/r2.1/api_docs/python/tf/estimator)
+    - [tf.keras](https://www.tensorflow.org/versions/r2.1/api_docs/python/tf/keras)
+
 
 ### Distributed training
 - [MirroredStrategy](https://www.tensorflow.org/versions/r1.15/api_docs/python/tf/distribute/MirroredStrategy) or [Contrib MirroredStrategy](https://www.tensorflow.org/versions/r1.15/api_docs/python/tf/contrib/distribute/MirroredStrategy)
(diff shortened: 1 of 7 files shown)


smdebug/tensorflow/collection.py
```diff
@@ -148,9 +148,6 @@ def __init__(self, collections=None, create_default=True):
                 self.create_collection(n)
             if is_tf_version_2x() and tf.executing_eagerly():
                 self.get(CollectionKeys.BIASES).include("^(?!gradient).*bias")
-                self.get(CollectionKeys.WEIGHTS).include("^weights/.*/((?!bias).)*$")
-                self.get(CollectionKeys.LOSSES).include(".*loss.*")
-                self.get(CollectionKeys.GRADIENTS).include("^gradient")
             else:
                 self.get(CollectionKeys.BIASES).include("bias")
```
vandanavk, line 152: why weren't loss and othernregexes set for SessionHook ?

docs/tensorflow.md
```diff
@@ -18,9 +18,14 @@
 - This library itself supports the following versions when you use our API which requires a few minimal changes to your training script: TensorFlow 1.14, 1.15, 2.0.1, 2.1.0. Keras 2.3.
```
vandanavk, line 18: maybe we just say we support 2.0+ versions

smdebug/tensorflow/collection.py
rahul003: Because SessionHook doesn't identify these tensors based on regex patterns. They come from global collections maintained by TF. (Not gradients where we manually add them)

rahul003: Why are the above being removed btw?

tests/zero_code_change/tf_utils.py
```diff
@@ -232,6 +231,8 @@ def neural_net(x):
 
 
 def get_data() -> "tf.contrib.learn.python.learn.datasets.base.Datasets":
+    from tensorflow.examples.tutorials.mnist import input_data
+
```
NihalHarish, line 235: Why?

smdebug/tensorflow/keras.py
```diff
@@ -716,6 +716,11 @@ def run(*args, **kwargs):
                 # at this point we need all collections to be ready
                 # this may not be the case at creation of hook
                 # as user's code after hook might add collections
+                self.collection_manager.get(CollectionKeys.WEIGHTS).include(
+                    "^weights/.*/((?!bias).)*$"
+                )
+                self.collection_manager.get(CollectionKeys.LOSSES).include(".*loss.*")
+                self.collection_manager.get(CollectionKeys.GRADIENTS).include("^gradient")
```
NihalHarish, line 723: It doesn't seem like we have tested the regex patterns themselves.

smdebug/tensorflow/collection.py
vandanavk: If I have loss regex here, then https://github.com/awslabs/sagemaker-debugger/blob/master/smdebug/tensorflow/session.py#L206 starts adding other operations in the graph with the regex "*loss*" into the loss collection. It gave 105 loss tensors in the trial instead of 1.

since these regex patterns were added for gradient tape support, moved them to grad tape related function so that it doesn't affect other hooks

tests/zero_code_change/tf_utils.py
vandanavk: its causing an error with TF 2 because I'm importing tf_utils in test_estimator.py.

smdebug/tensorflow/keras.py
vandanavk: copied them from pt regex. do we have tests for regex?

tests/zero_code_change/tf_utils.py
NihalHarish: Is there a function in this file that is exclusively used with TF 2?
Then we can simply move it to a new utils file, so that we do not have this import clash.

vandanavk: there's nothing specific to TF 2. whatever is used in the TF 2 works for both TF 1 and TF 2. only this get data function is specific to TF 1

Issue #261, opened 2020-05-28
Too many warnings printed with TF 2.X


vandanavk:
While executing tests that involve reductions, too many warnings are being printed.

Example
```

tests/tensorflow2/test_keras_mirrored.py::test_base_reductions INFO:absl:Load dataset info from /root/tensorflow_datasets/mnist/3.0.1
--
587 | INFO:absl:Field info.citation from disk and from code do not match. Keeping the one from code.
588 | INFO:absl:Reusing dataset mnist (/root/tensorflow_datasets/mnist/3.0.1)
589 | INFO:absl:Constructing tf.data.Dataset for split None, from /root/tensorflow_datasets/mnist/3.0.1
590 | INFO:tensorflow:Using MirroredStrategy with devices ('/job:localhost/replica:0/task:0/device:GPU:0', '/job:localhost/replica:0/task:0/device:GPU:1', '/job:localhost/replica:0/task:0/device:GPU:2', '/job:localhost/replica:0/task:0/device:GPU:3')
591 | INFO:tensorflow:Using MirroredStrategy with devices ('/job:localhost/replica:0/task:0/device:GPU:0', '/job:localhost/replica:0/task:0/device:GPU:1', '/job:localhost/replica:0/task:0/device:GPU:2', '/job:localhost/replica:0/task:0/device:GPU:3')
592 | [2020-05-27 07:15:21.755 ae4bb918ce14:10191 INFO hook.py:191] tensorboard_dir has not been set for the hook. SMDebug will not be exporting tensorboard summaries.
593 | [2020-05-27 07:15:21.755 ae4bb918ce14:10191 INFO hook.py:236] Saving to /tmp/test
594 | [2020-05-27 07:15:21.755 ae4bb918ce14:10191 INFO state_store.py:67] The checkpoint config file /opt/ml/input/config/checkpointconfig.json does not exist.
595 | [2020-05-27 07:15:22.897 ae4bb918ce14:10191 INFO keras.py:68] Executing in TF2.x eager mode.SageMaker Debugger will not be saving gradients
596 | [2020-05-27 07:15:23.037 ae4bb918ce14:10191 INFO hook.py:372] Monitoring the collections: metrics, sm_metrics, biases, losses, weights
597 | INFO:tensorflow:batch_all_reduce: 6 all-reduces with algorithm = nccl, num_packs = 1, agg_small_grads_max_bytes = 0 and agg_small_grads_max_group = 10
[...]

PR #271, opened 2020-06-05
Remove Reduction Warnings

NihalHarish:
### Description of changes:
- Reduction values were obtained by passing values to `np.linalg.norm`. This fn however, does not support scalars and throws an error.
- The PR converts scalars to array-like objects before passing them to the fn.
- The `get_basic_numpy_reduction ` performed this check `if abs:`, which always evaluated to `True` because it clashed with the name of a python builtin.

#### Style and formatting:

#261

### DIFF
--- a/smdebug/core/reductions.py
+++ b/smdebug/core/reductions.py
@@ -29,14 +29,15 @@ def get_basic_numpy_reduction(reduction_name, numpy_data):
             return getattr(np, reduction_name)(numpy_data)
     elif reduction_name in ALLOWED_NORMS:
         if reduction_name in ["l1", "l2"]:
-            ord = int(reduction_name[1])
+            order = int(reduction_name[1])
         else:
-            ord = None
+            order = None
 
-        if abs:
-            rv = np.linalg.norm(np.absolute(numpy_data), ord=ord)
-        else:
-            rv = np.linalg.norm(numpy_data, ord=ord)
+        if np.isscalar(numpy_data):
+            # np.linalg.norm expects array-like inputs
+            # but numpy_data can sometimes be a scalar value
+            numpy_data = [numpy_data]
+        rv = np.linalg.norm(numpy_data, ord=order)
         return rv
     return None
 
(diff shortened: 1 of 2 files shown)


smdebug/core/reductions.py
```diff
@@ -29,14 +29,15 @@ def get_basic_numpy_reduction(reduction_name, numpy_data):
             return getattr(np, reduction_name)(numpy_data)
     elif reduction_name in ALLOWED_NORMS:
         if reduction_name in ["l1", "l2"]:
-            ord = int(reduction_name[1])
+            order = int(reduction_name[1])
```
NihalHarish, line 32: renamed this variable, since `ord` conflicts with the python built-in

smdebug/tensorflow/base_hook.py
```diff
@@ -527,6 +527,8 @@ def _make_numpy_array(tensor_value):
 
     @staticmethod
     def _get_reduction_of_data(reduction_name, tensor_value, tensor_name, abs):
+        if hasattr(tensor_value, "numpy"):
+            tensor_value = tensor_value.numpy()
```
NihalHarish, line 531: This step, extracts the numpy value from EagerTensors

vandanavk: how was it working before this change?

NihalHarish: Because of the incorrect `if abs:` check in the next fn call, we were always calling `np.absolute(numpy_data)`

vandanavk:
Looks like there are still a few of these warnings in tf2 build. can you check if they are genuine or due to a bug? 

its not as many as before, but there are 12 occurrences.

NihalHarish:
> Looks like there are still a few of these warnings in tf2 build. can you check if they are genuine or due to a bug?
> 
> it's not as many as before, but there are 12 occurrences in tf2 and 20 occurrences in tf1. these are for reduction of weights and gradients.

```
Could not compute reduction l1 of conv2d/weights/conv2d/kernel:0 due to Improper number of dimensions to norm.
...
Could not compute reduction l1 of conv2d/weights/conv2d/kernel:0 due to Improper number of dimensions to norm.
``` 
is caused because the dimensions of those numpy arrays are not 1 or 2 ( which is expected for l1 and l2 reductions)

PR #273, opened 2020-06-12
Revert "TF 2.x: Support for keras to estimator"

NihalHarish:
Reverts awslabs/sagemaker-debugger#268


rahul003:
Why? Did it introduce some problem?

NihalHarish:
CI is failing.

vandanavk:
> Why? Did it introduce some problem?

seems to be failing on GPU. tensor count mismatch. let me check if the fix is quick instead of reverting

Issue #296, opened 2020-07-29
Move Dataset to S3


NihalHarish:
https://archive.ics.uci.edu/ml/machine-learning-databases/iris/iris.data into /root/tensorflow_datasets/downloads/

PR #298, opened 2020-07-30
Load IRIS Dataset from S3

NihalHarish:
### Description of changes:
- Removes the dependency on the public dataset with an s3 bucket hosted by smdebug

#### Style and formatting:

#296 

### DIFF
--- a/tests/constants.py
+++ b/tests/constants.py
@@ -1,0 +1,1 @@
+TF_IRIS_DATASET_S3_PATH = "s3://smdebug-testing/datasets/"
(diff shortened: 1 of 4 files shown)


tests/constants.py
```diff
@@ -0,0 +1 @@
+TF_IRIS_DATASET_S3_PATH = "s3://smdebug-testing/datasets/"
```
vandanavk, line 1: this path is not specific to Iris for TF, so maybe drop the prefix TF_IRIS

NihalHarish: Yes

PR #300, opened 2020-07-31
Remove Problematic Log

NihalHarish:
### Description of changes:
- Redundant log, since gradients are now saved by smdebug
- Also throws `ValueError: I/O operation on closed file.` when run with pytest which seems to crash the DLC pipeline as well

#### Style and formatting:


Vikas-kum:
why is this happening?
>> Also throws ValueError: I/O operation on closed file. when run with pytest which seems to crash the DLC pipeline as well ?

PR #310, opened 2020-08-05
Bug fix for state_store behavior while checking for metadata.json file

anirudhacharya:
### Description of changes:

While training a model using MXnet estimator, when Sagemaker tries to upload checkpoints the following error is encountered:

```
File "/opt/ml/code/translation.py", line 224, in __call__
return super(NMTModel, self).__call__(src_seq, tgt_seq, src_valid_length, tgt_valid_length)
File "/usr/local/lib/python3.6/site-packages/mxnet/gluon/block.py", line 756, in __call__
hook(self, args)
File "/usr/local/lib/python3.6/site-packages/smdebug/mxnet/hook.py", line 143, in forward_pre_hook
self._increment_step()
File "/usr/local/lib/python3.6/site-packages/smdebug/core/hook.py", line 511, in _increment_step
self._write_state()
File "/usr/local/lib/python3.6/site-packages/smdebug/core/hook.py", line 523, in _write_state
if self.state_store.is_checkpoint_updated():
File "/usr/local/lib/python3.6/site-packages/smdebug/core/state_store.py", line 112, in is_checkpoint_updated
cp_file_sizes = [os.path.getsize(file) for file in checkpoint_files]
File "/usr/local/lib/python3.6/site-packages/smdebug/core/state_store.py", line 112, in <listcomp>
cp_file_sizes = [os.path.getsize(file) for file in checkpoint_files]
File "/usr/local/lib/python3.6/genericpath.py", line 50, in getsize
return os.stat(filename).st_size
FileNotFoundError: [Errno 2] No such file or directory: '/opt/ml/checkpoints/metadata.json.sagemaker-uploading'
[2020-07-31 05:38:53.405 ip-10-2-236-41.ec2.internal:144 INFO utils.py:25] The end of training job file will not be written for jobs running under SageMaker.
2020-07-31 05:38:54,756 sagemaker-training-toolkit ERROR ExecuteUserScriptError:
```

- This PR fixes the issue by accounting for the behavior of SageMaker's S3 Uploader.

Some more comments on this bug - 
 - This bug seems like an extreme corner case, and here is what is happening 
    - smdebug's `state_store` periodically checks the checkpoint directory to see if any new checkpoint files have been created so that it can upload it an S3 bucket. But the checkpoint directory also contains the `metadata.json` file; to ensure that the `state_store` only reads checkpoint files and not metadata files, we exclude the metadata file [here](https://github.com/awslabs/sagemaker-debugger/blob/master/smdebug/core/state_store.py#L29)
    - But this bug arises because of the behavior of the SageMaker's S3 uploader. When SageMaker uploads checkpoint files to the S3 bucket it temporarily renames `metadata.json` to `metadata.json.sagemaker-uploaded`. This new file exists only for a short time duration. But if the `state_store` makes [the file check](https://github.com/awslabs/sagemaker-debugger/blob/master/smdebug/core/state_store.py#L29-L30) within that period of time,
        ```python
        if file != METADATA_FILENAME:
            checkpoint_files.append(os.path.join(child, file))
        ```
        then this will cause an error and throw the following exception - 
        ```
        FileNotFoundError: [Errno 2] No such file or directory: '/opt/ml/checkpoints/metadata.json.sagemaker-uploading'
        ```

 - This error was probably introduced when me and Vikas made this change - https://github.com/awslabs/sagemaker-debugger/pull/209 and https://github.com/awslabs/sagemaker-debugger/pull/200 some time back. We did not catch this bug early on because we were not aware of SageMaker S3 Uploader's behavior of temporarily renaming a file to `metadata.json.sagemaker-uploading`

#### Style and formatting:


anirudhacharya:
we will wait for a fix confirmation from here before merging this - https://t.corp.amazon.com/P38273888/communication

anirudhacharya:
Some more comments on this bug - 
 - This bug seems like an extreme corner case, and here is what is happening 
    - smdebug's `state_store` periodically checks the checkpoint directory to see if any new checkpoint files have been created. But the checkpoint directory also contains the `metadata.json`; to ensure that the `state_store` only reads checkpoint files and not metadata files, we exclude the metadata file [here](https://github.com/awslabs/sagemaker-debugger/blob/master/smdebug/core/state_store.py#L29)
    - But the bug arises because of the behavior of the SageMaker's S3 uploader. When SageMaker uploads checkpoint files to the S3 bucket it temporarily renames `metadata.json` to `metadata.json.sagemaker-uploaded`. This new file exists only for a short time duration. But if the `state_store` makes [this](https://github.com/awslabs/sagemaker-debugger/blob/master/smdebug/core/state_store.py#L29-L30) file check 
        ```python
        if file != METADATA_FILENAME:
            checkpoint_files.append(os.path.join(child, file))
        ```
        within that period of time, then this will cause an error and throw the following exception - 
        ```
        FileNotFoundError: [Errno 2] No such file or directory: '/opt/ml/checkpoints/metadata.json.sagemaker-uploading'
        ```

 - This error was probably introduced when me and Vikas made this change - https://github.com/awslabs/sagemaker-debugger/pull/209 and https://github.com/awslabs/sagemaker-debugger/pull/200 some time back. We did not catch this bug early on because we were not aware of SageMaker S3 Uploader's behavior of temporarily renaming a file to `metadata.json.sagemaker-uploading`

Vikas-kum:
A simple deterministic fix would be arap os.path.getsize in try catch block . 
In a way look at all the places where the files are accessed and can be deleted modified by another process agent, it should all be handled by try catch, repeat block

smdebug/core/state_store.py
```diff
@@ -26,7 +27,7 @@ def _get_checkpoint_files_in_dir(self, cp_dir):
         checkpoint_files = []
         for child, _, files in os.walk(cp_dir):
             for file in files:
-                if file != METADATA_FILENAME:
+                if file != METADATA_FILENAME and file != METADATA_FILENAME_S3_UPLOADED:
```
Vikas-kum, line 30: this is not a right fix, 
you need to handle exception every where we are trying to access file ( The files can be deleted or modeified by a different process) . Assume that files can be concurrently modified by another process.

anirudhacharya: done

Vikas-kum, line 30: remove this not needed.

smdebug/core/state_store.py
```diff
@@ -40,12 +41,15 @@ def __init__(self):
...
-                os.path.getsize(file) for file in self._last_seen_checkpoint_files
-            ]
+            try:
+                self._last_seen_checkpoint_files = self._get_checkpoint_files_in_dir(
+                    self._checkpoint_dir
+                )
+                self._last_seen_cp_files_size = [
+                    os.path.getsize(file) for file in self._last_seen_checkpoint_files
+                ]
+            except FileNotFoundError as e:
```
Vikas-kum, line 51: except Exception , 
ideally we dont want to cause Training to fail

smdebug/core/state_store.py
```diff
@@ -40,12 +41,15 @@ def __init__(self):
...
+                logger.warning(e)
```
Vikas-kum, line 52: self._last_seen_checkpoint_files = [] 
self._last_seen_cp_files_size = []

Vikas-kum (commented):
Confirm if metadata file can be deleted by uploader after uploading. In that case we need to create a tmp file and always create a copy, I am hoping .tmp suffix wont be deleted.

Actually please create a file with .tmp extension and then rename it.

smdebug/core/state_store.py
```diff
@@ -96,20 +100,31 @@ def is_checkpoint_updated(self):
                 )
                 return False
             checkpoint_files = sorted(checkpoint_files)
-            timestamps = [os.path.getmtime(file) for file in checkpoint_files]
+            try:
+                timestamps = [os.path.getmtime(file) for file in checkpoint_files]
```
Vikas-kum, line 104: for loop :
  try: 
     timestamps[i] = 
  except:
    timestamps[i]=0

smdebug/core/state_store.py
```diff
@@ -96,20 +100,31 @@ def is_checkpoint_updated(self):
...
             )
 
             if len(self._last_seen_checkpoint_files) != len(checkpoint_files):
                 self._last_seen_checkpoint_files = checkpoint_files
-                self._last_seen_cp_files_size = [os.path.getsize(file) for file in checkpoint_files]
+                try:
+                    self._last_seen_cp_files_size = [
+                        os.path.getsize(file) for file in checkpoint_files
+                    ]
+                except FileNotFoundError as e:
```
Vikas-kum, line 117: except Exception, here and all other places.

anirudhacharya:
> Confirm if metadata file can be deleted by uploader after uploading. In that case we need to create a tmp file and always create a copy, I am hoping .tmp suffix wont be deleted.

how do i do this?

smdebug/core/state_store.py
anirudhacharya: why do we want to empty these two lists here?

smdebug/core/state_store.py
anirudhacharya: sure.

anirudhacharya: removing this fails this test - https://github.com/awslabs/sagemaker-debugger/pull/310/files#diff-9ec1a0fff79c490c65c2ae46f7e51d09R39

Vikas-kum:
> > Confirm if metadata file can be deleted by uploader after uploading. In that case we need to create a tmp file and always create a copy, I am hoping .tmp suffix wont be deleted.
> 
> how do i do this?( more specifically this is not something i can do locally right?)

Confirm the bhavior with uploader team. You need to take care of - 
1) If uploader process can remove the file 
2) What happens when you find the file that is deleted. Right now you are appending the file size to 0 , what will be the consequences of that change? Will smdebug treat it as new checkpoint and write new entry in metadata. Will that cause more frequent writes, etc.

anirudhacharya:
@Vikas-kum After talking to the uploader team, a few clarifications -
>
> Confirm the bhavior with uploader team. You need to take care of -
"there seems to be some confusion in that PR comm - the XXX.sagemaker-upload(ing|ed) files are flag files, they are not renamed files - they signify the states of the upload process. 

The uploader makes no distinction between files, it knows nothing about your metadata file AFAIK, the contract is for it to upload files it finds in the configured continuous-upload directory and then delete it once the file once the upload is completed. If you do not want a particular file to be subject to this treatment you shouldn’t be placing it in that same dir"

> 
> 1. If uploader process can remove the file
> 2. What happens when you find the file that is deleted. Right now you are appending the file size to 0 , what will be the consequences of that change? Will smdebug treat it as new checkpoint and write new entry in metadata. Will that cause more frequent writes, etc.

I will re-evaluate this.

setu4993:
Is there an update on this and / or when this will be merged? AWS SageMaker customer here, and I can confirm that I'm seeing this error pop up.

setu4993:
Is there also a recommended workaround for this in the meanwhile? A custom hook config?

anirudhacharya:
> Is there also a recommended workaround for this in the meanwhile? A custom hook config?

Workaround is to pass this `debugger_hook_config=False` in the estimator function.

anirudhacharya:
> Is there an update on this and / or when this will be merged? AWS SageMaker customer here, and I can confirm that I'm seeing this error pop up.

ETA - next week.

setu4993:
Thanks @anirudhacharya! Setting `debugger_hook_config=False` might not be ideal... I think it'll disable checkpoint uploads during training for spot instances.

setu4993:
Never mind, can still get checkpoints during a spot training job. Thanks for the help!

rahul003:
Seems to have been fixed in https://github.com/awslabs/sagemaker-debugger/pull/404

PR #320, opened 2020-08-06
Update Assert Statements For New TF 2.2.0 DLC

NihalHarish:
### Description of changes:
- Update asserts to work with the latest DLC

#### Style and formatting:


tests/zero_code_change/test_tensorflow2_gradtape_integration.py
```diff
@@ -26,10 +25,13 @@ def get_keras_data():
...
 
 
-def helper_test_keras_v2_gradienttape(script_mode: bool = False, json_file_contents="{}"):
+def helper_test_keras_v2_gradienttape(
+    script_mode: bool = False, json_file_contents="{}", default=False
+):
     """ Test the default ZCC behavior of saving losses and metrics in eager and non-eager modes."""
     smd.del_hook()
     tf.keras.backend.clear_session()
+    is_tf_2_2 = True if tf.__version__ == "2.2.0" else False
```
vandanavk, line 34: can we use the test util?

NihalHarish: We can. I have reverted it

NihalHarish:
> relax the asserts to >= so that the tests dont fail on vanilla

Since we do not currently test on a vanilla TF 2x container, I'm keeping the strict asserts

tests/zero_code_change/test_tensorflow2_gradtape_integration.py
NihalHarish: I have reverted this.

Issue #321, opened 2020-08-07
Training crashes with DebuggerHookConfig


ratulray:
I have a simple Tensorflow model which I'm training on SageMaker. 
It was working fine. Recently it has started crashing while training **right after the first checkpoint is saved**.

Following is the summary of the trace:

`AssertionError: training/Ftrl/Ftrl/AssignAddVariableOp is not in graph`
```
Traceback (most recent call last):
  File "trainer_click_label.py", line 259, in <module>
    estimator = train_and_evaluate(train_files, test_files, args.model_dir, args.epochs)
  File "trainer_click_label.py", line 133, in train_and_evaluate
    tf.estimator.train_and_evaluate(linear_est, train_spec=train_spec, eval_spec=eval_spec)
  File "/usr/local/lib/python3.6/dist-packages/tensorflow_estimator/python/estimator/training.py", line 473, in train_and_evaluate
    return executor.run()
  File "/usr/local/lib/python3.6/dist-packages/tensorflow_estimator/python/estimator/training.py", line 613, in run
    return self.run_local()
  File "/usr/local/lib/python3.6/dist-packages/tensorflow_estimator/python/estimator/training.py", line 714, in run_local
    saving_listeners=saving_listeners)
  File "/usr/local/lib/python3.6/dist-packages/tensorflow_estimator/python/estimator/estimator.py", line 374, in train
    loss = self._train_model(input_fn, hooks, saving_listeners)
  File "/usr/local/lib/python3.6/dist-packages/tensorflow_estimator/python/estimator/estimator.py", line 1164, in _train_model
    return self._train_model_default(input_fn, hooks, saving_listeners)
  File "/usr/local/lib/python3.6/dist-packages/tensorflow_estimator/python/estimator/estimator.py", line 1198, in _train_model_default
    saving_listeners)
  File "/usr/local/lib/python3.6/dist-packages/tensorflow_estimator/python/estimator/estimator.py", line 1497, in _train_with_estimator_spec
    _, loss = mon_sess.run([estimator_spec.train_op, estimator_spec.loss])
  File "/usr/local/lib/python3.6/dist-packages/tensorflow_core/python/training/monitored_session.py", line 784, in run
    run_metadata=run_metadata)
  File "/usr/local/lib/python3.6/dist-packages/tensorflow_core/python/training/monitored_session.py", line 1289, in run
    run_metadata=run_metadata)
[...]

PR #324, opened 2020-08-11
Adding filtering node logic issue:321

Vikas-kum:
Issue: https://github.com/awslabs/sagemaker-debugger/issues/321

### Description of changes:
Adding filtering logic to dest_names and make sure that we always ask for subgraph which is present in graph def 

#### Style and formatting:

### DIFF
--- a/smdebug/tensorflow/collection.py
+++ b/smdebug/tensorflow/collection.py
@@ -18,13 +18,13 @@ logger = get_logger()
 
 class Collection(BaseCollection):
     def __init__(
-            self,
-            name,
-            include_regex=None,
-            tensor_names=None,
-            reduction_config=None,
-            save_config=None,
-            save_histogram=True,
+        self,
+        name,
+        include_regex=None,
+        tensor_names=None,
+        reduction_config=None,
+        save_config=None,
+        save_histogram=True,
     ):
         super().__init__(
             name, include_regex, tensor_names, reduction_config, save_config, save_histogram
@@ -148,7 +148,9 @@ class Collection(BaseCollection):
             self._graph_tensors_map[graph] = {}
         if name not in self._graph_tensors_map[graph]:
             self._graph_tensors_map[graph][name] = tensor
-            get_logger().debug(f"Added name:{name} to graph:{graph} to tensor:{tensor} self:{id(self)}")
+            get_logger().debug(
+                f"Added name:{name} to graph:{graph} to tensor:{tensor} self:{id(self)}"
+            )
 
         self.add_tensor_name(export_name)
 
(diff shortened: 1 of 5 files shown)


smdebug/tensorflow/collection.py
```diff
@@ -73,8 +73,10 @@ def add_for_mode(self, arg, mode=None):
          Adds tensors to the collection from a given Operation, Tensor, Variable or MirroredVariable
          :param arg: the argument to add to collection
          """
+        get_logger().debug(f"type:{type(arg)} arg:{arg} self:{id(self)}")
```
rahul003, line 76: Did you intend to remove before PR? If not, you might want to add some description before.

smdebug/tensorflow/collection.py
```diff
@@ -124,11 +129,18 @@ def set_tensor_ref(self, tensor, tensor_name: tf.Tensor = None):
...
         self.add_tensor_name(export_name)
 
-    def has_tensor(self, name):
-        # tf object name
+        if name != export_name:
+            get_logger().debug(
+                f"Export_name:{export_name} != name:{name} . Adding export_name:{export_name} to include in "
+                f"collection collection_name:{self.name} selfId:{id(self)} "
+            )
+        self.include("^" + export_name + "$")
```
rahul003, line 141: This doesn't look correct. name is supposed to be the the actual name of tensor from TF's point of view. export_name is a more meaningful name for that tensor for user's benefit. Including export name is not useful and in fact might be wrong in some situations if that refers to a different tensor.

smdebug/tensorflow/collection.py
```diff
@@ -108,7 +110,10 @@ def get_export_names_of_tensors(self):
         return self.tensor_names
 
     def get_tensor(self, name):
-        return self._tensors[name]
+        if name in self._tensors:
+            return self._tensors[name]
+        else:
+            return None
```
rahul003, line 116: Could you check if returning None causes issues wherever this function is used? I don't think None is expected. Is it actually possible? Did you see it happening?

smdebug/tensorflow/collection.py
Vikas-kum: do we use export name in case of keras or for other cases also?
keras regex match works on export_name i guess.

rahul003: export name was only used for Keras AFAIK. @NihalHarish

rahul003: Why did you need this though?

Vikas-kum: for Tf estimator , graph is different for train and eval phase. Loss tensors are getting collected from loss collecection. 
During eval phase loss collection is empty. So we don't save any tensor. 

This change ensures that if we are saving some tensor, we add the name of the tensor being saved as regex, so that we always save these tensors if name is part of graph.

Vikas-kum: So, question is if export_name is there, is it used for searching collection regex? 
In other words - 
in what cases tensor_name won't be there passed to this function and 
what(name or export name?) is being used for searching collection regex in these cases ?

PR #328, opened 2020-08-14
Add ability to only save shapes of tensors

rahul003:
### Description of changes:
- Added a config to ReductionConfig to save shape. 
- Created a class ShapeWriter which uses the same index writer as (Event)FileWriter and adds the shapes to the index file. Adding shapes to the index file means that we get index files even when event files doesn't exist. It helps to not create more files though.
- Added an API `shape` to tensor class.
- Added tests for mxnet, tensorflow, pytorch

#### Style and formatting:

### DIFF
--- a/smdebug/core/index_reader.py
+++ b/smdebug/core/index_reader.py
@@ -241,8 +241,9 @@ class IndexReader(ABC):
         else:
             event_file_name = None
 
-        tensor_payload = index_dict["tensor_payload"]
         to_update_index_dict = []
+
+        tensor_payload = index_dict.get("tensor_payload", [])
         for tensor in tensor_payload:
             tensor_name = tensor["tensorname"]
             start_idx = tensor["start_idx"]
@@ -252,7 +253,7 @@ class IndexReader(ABC):
             )
             to_update_index_dict.append((tensor_name, step, tensor_location))
 
-        shape_payload = index_dict["shape_payload"]
+        shape_payload = index_dict.get("shape_payload", [])
         for tensor in shape_payload:
             tensor_name = tensor["tensorname"]
             original_name = tensor["originalname"]
--- a/smdebug/core/index_reader.py
+++ b/smdebug/core/index_reader.py
@@ -248,7 +248,7 @@ class IndexReader(ABC):
 
         to_update_index_dict = []
 
-        if len(index_dict["tensor_payload"]):
+        if "tensor_payload" in index_dict and len(index_dict["tensor_payload"]):
             event_file_name = os.path.join(path, index_meta["event_file_name"])
             for tensor in index_dict["tensor_payload"]:
                 tensor_name = tensor["tensorname"]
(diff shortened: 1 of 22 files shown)


rahul003:
I don't have access to the CI account. What's the error?

NihalHarish:
> ### Description of changes:
> * Added a config to ReductionConfig to save shape.
> * Created a class ShapeWriter which uses the same index writer as (Event)FileWriter and adds the shapes to the index file.
> * Added an API `shape` to tensor class.
> * Added tests for mxnet, tensorflow, pytorch
> 
> #### Style and formatting:
> I have run `pre-commit install` to ensure that auto-formatting happens with every commit.
> 
> #### Issue number, if available
> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Can you explain like we discussed earlier, why we're logging shapes to the index files and having them as independent files or saving them to event files.

smdebug/core/index_reader.py
```diff
@@ -233,28 +235,45 @@ def _update_tensors_from_json(
...
+
+        to_update_index_dict = []
+
+        tensor_payload = index_dict.get("tensor_payload", [])
+        for tensor in tensor_payload:
             tensor_name = tensor["tensorname"]
             start_idx = tensor["start_idx"]
             length = tensor["length"]
             tensor_location = TensorLocation(
                 tensor_name, mode, mode_step, event_file_name, start_idx, length, worker
```
NihalHarish, line 252: Can `event_file_name` be `None` here?

`if event_file_name in index_meta is False` and `len(tensor_payload)>0`.

Is the TensorLocation designed to handle this?

Is there a possibility for the user to attempt to access  the tensor value when it is None?

rahul003: Previously the expectation was that event_file_name is always present when tensor_payload is in the index file. But good point it could have led to silent failures or issues further down the line. I've refactored the code. Now if event_file_name is not present when reading tensor_payload, it'll throw KeyError

rahul003: > Is there a possibility for the user to attempt to access the tensor value when it is None but there is corresponding shape payload

That's very much possible and would raise TensorUnavailableForStep exception.

rahul003:
How to get codecov to rerun?

rahul003:
> > ### Description of changes:
> > 
> > * Added a config to ReductionConfig to save shape.
> > * Created a class ShapeWriter which uses the same index writer as (Event)FileWriter and adds the shapes to the index file.
> > * Added an API `shape` to tensor class.
> > * Added tests for mxnet, tensorflow, pytorch
> > 
> > #### Style and formatting:
> > I have run `pre-commit install` to ensure that auto-formatting happens with every commit.
> > #### Issue number, if available
> > By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.
> 
> Can you explain like we discussed earlier, why we're logging shapes to the index files and having them as independent files or saving them to event files.

I didn't follow this, but I have updated the description. Does that answer your question?

PR #330, opened 2020-08-18
Add Support For More TF Types

NihalHarish:
### Description of changes:
- Add support for any TF Type that has a TF to Numpy mapping
- Add a test to check if the mapping dict becomes stale
- A bug that was recently reported was because the assumption that tf tensors are mapped to a single numpy type is wrong.
For example, tf string can have `npdtype.kind == "U" or npdtype.kind == "O" or npdtype.kind == "S"` ( unicode, object, str).

- The user does not control this low level detail and is determined by factors like training mode, dataset et cetera.
- This PR introduces a test that is run with the `_NP_TO_TF` datastructure from `tensorflow.python.framework.dtypes` 
- It will break if a new numpy datatype support is added, prompting the smdebug project to also sync
#### Style and formatting:


smdebug/core/tfevent/util.py
```diff
@@ -10,6 +11,9 @@
 
 logger = get_logger()
 
+# TF Implements a Custom Numpy Datatype for Brain Floating Type
+_np_bfloat16 = _pywrap_bfloat16.TF_bfloat16_type()
```
vandanavk, line 15: is there a specific requirement for this?

tests/tensorflow2/test_tensorflow2_datatypes.py
```diff
@@ -0,0 +1,16 @@
+# Third Party
+from tensorflow.python.framework.dtypes import _NP_TO_TF
+
+# First Party
+from smdebug.core.tfevent.util import _get_proto_dtype
+
+
+def test_tensorflow2_datatypes():
```
vandanavk, line 8: apart from this unit test, is there a training script that makes use of these different datatypes, that we can use for testing?

NihalHarish: The this test tracks all possible TF type to numpy types.

Any update to the mapping will make this test break, effectively helping us track possible tf to numpy type conversation.

smdebug/core/tfevent/util.py
```diff
@@ -27,12 +27,25 @@
...
 
 def _get_proto_dtype(npdtype):
-    if npdtype.kind == "U":
+    if npdtype.kind == "U" or npdtype.kind == "O" or npdtype.kind == "S":
         return (False, "DT_STRING")
+    try:
+        from tensorflow.python import _pywrap_bfloat16
+
+        # TF Implements a Custom Numpy Datatype for Brain Floating Type
+        _np_bfloat16 = _pywrap_bfloat16.TF_bfloat16_type()
```
NihalHarish, line 45: There is no special ask for this, but the goal is to support every possible TF type.

smdebug/core/tfevent/util.py
NihalHarish: There is no special ask for this, but the goal is to support every possible TF type.

tests/tensorflow2/test_tensorflow2_datatypes.py
Vikas-kum: I guess we should also test, if these new types tensors are saved they are saved without errors. 
You should create tensors and may be try using save_tensor API

tests/tensorflow2/test_tensorflow2_datatypes.py
```diff
@@ -0,0 +1,17 @@
+# Third Party
+import numpy as np
+from tensorflow.python.framework.dtypes import _NP_TO_TF
+
+# First Party
+from smdebug.core.tfevent.util import _get_proto_dtype
+
+
+def test_tensorflow2_datatypes():
```
vandanavk, line 9: add this test to TF1 as well. since common code is being changed

PR #332, opened 2020-08-18
Function to Test If the hook has been configured with the Default hook config

NihalHarish:
### Description of changes:
- Defines a function that checks if the hook is running with the default hook configuration
- This is especially needed in AWS TF code, to turn off certain features like capturing outputs if the hook hasn't been configured to do so

#### Style and formatting:


smdebug/core/config_constants.py
```diff
@@ -41,3 +41,5 @@
 
 CALLABLE_CACHE_ENV_VAR = "SMDEBUG_KERAS_CALLABLE_CACHE_TYPE"
 DEFAULT_CALLABLE_CACHE = "CACHE_PER_MODE"
+
+DEFAULT_SAVED_COLLECTIONS = ["losses", "metrics", "sm_metrics"]
```
vandanavk, line 45: AFAIK metrics are relevant only to TF. will this fail for MX and PT?

smdebug/core/hook.py
```diff
@@ -364,6 +365,14 @@ def _get_collections_with_tensor(self, tensor_name) -> Set["Collection"]:
     def _get_default_collections(self):
         pass
 
+    def has_default_hook_configuration(self):
```
vandanavk, line 368: this will be called by framework code I assume?

smdebug/core/config_constants.py
NihalHarish: Okay, didn't know that.
And the tests failed for MxNet and Pytorch

smdebug/core/hook.py
NihalHarish: Yes, I have added comments.

NihalHarish:
> > > This is especially needed in AWS TF code, to turn off certain features like capturing outputs if the hook hasn't been configured to do so
> 
> Will this break if I add one more collection
> MyCollection with tensor 'a' ?

Yes. If the hook is not executing with the default hook configuration we will capture values but not save ( write them )

Vikas-kum:
> > > > This is especially needed in AWS TF code, to turn off certain features like capturing outputs if the hook hasn't been configured to do so
> > 
> > 
> > Will this break if I add one more collection
> > MyCollection with tensor 'a' ?
> 
> Yes. If the hook is not executing with the default hook configuration we will capture values but not save ( write them )

Will that again cause regression ? 
Why do we have to even capture the tensor until and unless it is asked for ?

NihalHarish:
> > > > > This is especially needed in AWS TF code, to turn off certain features like capturing outputs if the hook hasn't been configured to do so
> > > 
> > > 
> > > Will this break if I add one more collection
> > > MyCollection with tensor 'a' ?
> > 
> > 
> > Yes. If the hook is not executing with the default hook configuration we will capture values but not save ( write them )
> 
> Will that again cause regression ?
> I guess we should not even capture the tensor until and unless it is asked for ?

Let's pause this discussion.
I will deliver a perf doc that tracks performance regressions with different collection configs.

smdebug/tensorflow/keras.py
```diff
@@ -348,7 +349,10 @@ def _prepare_tensors_available_post_step(self):
 
                 # Add tensor to custom collections
                 for custom_coll in custom_collections:
-                    if match_inc(tensor_ref.name, custom_coll.include_regex):
+                    if (
+                        match_inc(tensor_ref.name, custom_coll.include_regex)
+                        and tensor_ref.tf_obj is not None
```
vandanavk, line 354: why this check? was there an issue without it?

smdebug/tensorflow/keras.py
```diff
@@ -390,6 +394,30 @@ def _save_custom_tensors_post_step(self):
...
+        # should be saved
+        if self._is_collection_being_saved_for_step(CollectionKeys.LAYERS):
+            return True
+        collections_to_save = self._get_collections_to_save_for_step()
+        for c in collections_to_save:
+            if match_inc(layer_name, c.include_regex):
+                return True
+        return False
+
+    def should_save_gradient(self, layer_name):
```
vandanavk, line 409: can you combine this with should_save_layer() ? the definition is the same except for the collection name

smdebug/tensorflow/keras.py
```diff
@@ -489,11 +531,8 @@ def _save_metrics(self, batch, logs, force_save=False):
                     self._add_metric(metric_name=key)
                     self._save_for_tensor(key, logs[key], check_before_write=False)
 
-    def _save_layer_input_and_outputs(self, grad_tape=False):
-        # Iterates over all the saved layers for input and output values
-        if is_tf_version_2x() is False or (grad_tape is False and self.model.run_eagerly is False):
-            # This function only works when the run_eagerly is True
-            return
+    def _save_layer_input_and_outputs(self):
+        # Run only for GradTape
```
vandanavk, line 535: if its only for gradtape, should you check if self.tape is True?

smdebug/tensorflow/keras.py
```diff
@@ -698,33 +736,19 @@ def on_test_batch_begin(self, batch, logs=None):
...
-    def save_layer_inputs(self, x, layer_outputs, model=None):
-        self._save_layer_values(
-            layer_outputs, self.get_collection(CollectionKeys.LAYERS), model, inputs=x
-        )
+    def _save_layer_values(self, logs):
+        if logs is None:
+            return
+        step_collections = self._get_collections_to_save_for_step()
+        layer_collection = self.get_collection(CollectionKeys.LAYERS)
+        collections_to_write = {layer_collection} if layer_collection in step_collections else set()
```
vandanavk, line 744: will we be looping through logs even if collections_to_write = set() ?

vandanavk: is this "if" equal to checking 

```
if should_save_layer():
    collections_to_write = {self.get_collection(CollectionKeys.LAYERS)}
```
?

tests/tensorflow/keras/test_keras.py
```diff
@@ -341,7 +341,7 @@ def test_include_regex(out_dir):
     tr = create_trial_fast_refresh(out_dir)
     tnames = tr.tensor_names(collection="custom_coll")
 
-    assert len(tnames) == 12
+    assert len(tnames) == 8
```
vandanavk, line 344: why this change?

tests/zero_code_change/test_tensorflow2_integration.py
```diff
@@ -117,6 +117,9 @@ def helper_test_keras_v2(script_mode: bool = False, eager_mode: bool = True):
 
         hook = smd.get_hook()
         assert hook
+        # Check if the hook was executed with the default
+        # hook configuration
+        assert hook.has_default_hook_configuration()
```
vandanavk, line 122: isn't this applicable to the gradtape integration test too?

smdebug/tensorflow/keras.py
NihalHarish: tensor_ref.tf_obj is none for certain tensors (kernelGrad) and ends up throwing a warning

smdebug/tensorflow/keras.py
NihalHarish: Sure

tests/tensorflow/keras/test_keras.py
NihalHarish: Because we stop passing gradients by default from TF to smdebug.

So there is no way to catch them by simply defining a regex

tests/zero_code_change/test_tensorflow2_integration.py
NihalHarish: Yes, will add the check there as welll

smdebug/tensorflow/keras.py
NihalHarish: We can reduce some duplication here for sure.
But the idea behind two functions was to eliminate the need to use strings or string-constants in AWS TF, since we do not have the CollectionKeys Enum defined in their codebase.

smdebug/tensorflow/keras.py
NihalHarish: We need to loop though logs if collections_to_write={} because:
- layers are passed to smdebug only if the user has requested for it with the LAYERS collection or regex.
- if CollectionKeys.LAYERS is not present then we must still save since it might be present in a regex based collection

NihalHarish: We'll be saving to custom collections with a matching regex

smdebug/tensorflow/keras.py
```diff
@@ -418,6 +438,34 @@ def _save_tensor_to_file(self, tensor_name, tensor_value, collections):
...
             self._save_for_tensor(tensor_name, t, check_before_write=True)
 
+    def save_gradients_from_logs(self, gradients):
+        if gradients is not None:
+            if len(self.gradient_name_to_tensor_position_map) == 0 and self._gradients_set is False:
+                # Map has not yet been initialized
+                for index, grad_var in enumerate(zip(gradients, self.model.trainable_variables)):
+                    g, v = grad_var  # split gradients and trainable variables tuple
+                    layer_name = v.name
+                    if len(layer_name.split(":")) > 1:
```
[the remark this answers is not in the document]
NihalHarish: if the character `:` is not present in the string, then split returns a single single string.
The next line will simply be a redundant operation.

```
>>> s = "helloworld"
>>> s.split(':')
['helloworld']
```

smdebug/tensorflow/keras.py
```diff
@@ -390,6 +394,22 @@ def _save_custom_tensors_post_step(self):
             self._save_tensor_to_file(tensor_name, tensor_value, collection_names)
         self.custom_tensors_to_save.clear()
 
+    def should_save_layer(self, layer_name):
+        # Called in AWS TF to determine
+        # if a particular layer value
+        # should be saved
+        if self._is_collection_being_saved_for_step(CollectionKeys.LAYERS):
+            return True
+        return self.is_tensor_in_custom_collection(layer_name)
```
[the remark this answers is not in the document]
NihalHarish: Yes. The function then returns False to the caller and the value of the layer isn't saved.

smdebug/tensorflow/keras.py
```diff
@@ -418,6 +438,34 @@ def _save_tensor_to_file(self, tensor_name, tensor_value, collections):
...
+            if len(self.gradient_name_to_tensor_position_map) == 0 and self._gradients_set is False:
+                # Map has not yet been initialized
+                for index, grad_var in enumerate(zip(gradients, self.model.trainable_variables)):
+                    g, v = grad_var  # split gradients and trainable variables tuple
+                    layer_name = v.name
+                    if len(layer_name.split(":")) > 1:
+                        layer_name = layer_name.split(":")[0]
+                    export_name = "gradients/" + layer_name + "Grad"
+                    if self.should_save_gradient(export_name):
+                        self.gradient_name_to_tensor_position_map[export_name] = index
```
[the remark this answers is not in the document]
NihalHarish: We currently save gradients only when the user explicitly sets the CollectionKey, so this is currently not possible

smdebug/tensorflow/keras.py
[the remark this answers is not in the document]
ndodda-amazon: Do we even need this check? As pointed out in your example, if `:` is not present, it will return the string wrapped in an array. So it looks like either way, `layer_name.split(":")[0]` is enough to get the expected value.
```
>>> s1 = "helloworld"
>>> s1.split(':')[0]
'helloworld'
>>> s2 = "hello:world"
>>> s2.split(':')[0]
'hello'
```

smdebug/tensorflow/keras.py
```diff
@@ -668,7 +705,7 @@ def _on_any_batch_begin(self, batch, mode, logs=None):
             if (is_tf_version_2x() and tf.executing_eagerly()) or self._validate_exec_function(
                 self._get_exec_function(mode)
             ):
-                self._wrap_model_with_input_output_saver()
+                # self._wrap_model_with_input_output_saver()
```
ndodda-amazon, line 708: nit: If this line is not needed, can it be removed?

NihalHarish: Will do. Thanks!

smdebug/tensorflow/keras.py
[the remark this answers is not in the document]
NihalHarish: You're right. It seems to work fine for empty strings as well.

smdebug/core/hook.py
```diff
@@ -552,6 +561,13 @@ def _increment_step(self):
             self.mode_steps[ModeKeys.GLOBAL] = self.step
         self._collections_to_save_for_step = None
 
+    def should_save_tensor(self, tensor_name: str, collection_name: str) -> bool:
+        # Called in the internal AWS codebase to determine
+        # if a particular tensor value should be saved
+        if self._is_collection_being_saved_for_step(collection_name):
+            return True
+        return self.is_tensor_in_custom_collection(tensor_name)
```
Vikas-kum, line 569: if collection is not saved for step , you just check is_tensor_in_custom_collection, shouldnt you check should_save_tensor_for_step ?

Also, what is the step val at this point ? Say this call is made before step 1, then what is the value of step right now in smdebug? 
Can there be off by one error?

smdebug/tensorflow/base_hook.py
```diff
@@ -218,6 +219,14 @@ def export_collections(self):
         collection_file_name = f"{self.worker}_collections.json"
         self.collection_manager.export(self.out_dir, collection_file_name)
 
+    def has_default_hook_configuration(self):
```
Vikas-kum, line 222: why this function is repeated again here?

smdebug/tensorflow/keras.py
```diff
@@ -109,7 +113,8 @@ def register_model(self, model):
         # It attaches a hook to every layer of the model to capture
         # layer values
         self.model = model
-        self._wrap_model_with_input_output_saver()
+        if self.tape is not None:
```
Vikas-kum, line 116: why this change?

smdebug/tensorflow/base_hook.py
NihalHarish: I have overridden the fn in core hook to utilize a constant defined for TF.

smdebug/tensorflow/keras.py
NihalHarish: The wrap model isn't needed with the new changes on TF to capture layer inputs and outputs when we use `.fit`

smdebug/core/hook.py
NihalHarish: The call cannot be made before the step is incremented because training only starts after the `on_train_begin callback` is executed.

smdebug/core/hook.py
```diff
@@ -343,9 +344,10 @@ def _get_collections_to_save_for_step(self) -> Set["Collection"]:
                 )
         return self._collections_to_save_for_step
 
-    def _saving_shapes_in_step(self) -> bool:
-        for coll in self._get_collections_to_save_for_step():
-            if coll.reduction_config.save_shape is True:
+    def is_tensor_saved_for_step(self, tensor_name):
```
Vikas-kum, line 347: should_save_tensor_for_step ?

smdebug/core/hook.py
```diff
@@ -525,6 +541,13 @@ def _increment_step(self):
             self.mode_steps[ModeKeys.GLOBAL] = self.step
         self._collections_to_save_for_step = None
 
+    def should_save_tensor(self, tensor_name: str, collection_name: str) -> bool:
```
Vikas-kum, line 544: should_save_tensor_or_collection

Vikas-kum: write comment at top of method.

NihalHarish, line 544: Rename to `should_save_tensor_for_collection`

smdebug/tensorflow/keras.py
```diff
@@ -71,6 +72,9 @@ def __init__(
         )  # stores tensors custom tensors saved by users every step
         self.saved_layers = dict()
         self.has_registered_model = False
+        self.gradient_name_to_tensor_position_map = dict()
```
NihalHarish, line 75: Delete

smdebug/tensorflow/keras.py
```diff
@@ -71,6 +72,9 @@ def __init__(
...
+        self._supports_tf_logs = True
```
NihalHarish, line 76: Comment

Vikas-kum, line 76: add a comment.

smdebug/tensorflow/keras.py
```diff
@@ -71,6 +72,9 @@ def __init__(
...
+        self.step_incremented_in_on_train_begin = False
```
Vikas-kum, line 77: comment about what is this.

PR #333, opened 2020-08-19
Save tf2 model

vandanavk:
### Description of changes:
Inputs to layers.__call__ are expected to be called "inputs" as self._call_fn_args is by default set to ["inputs"]. Failure to do so results in errors while trying to save the model.

#### Style and formatting:

### DIFF
--- a/smdebug/tensorflow/utils.py
+++ b/smdebug/tensorflow/utils.py
@@ -329,7 +329,7 @@ class InputOutputSaver:
         self.layer_input = None
         self.layer_output = None
 
-    def __call__(self, callable_inputs, *args, **kwargs) -> None:
+    def __call__(self, inputs, *args, **kwargs) -> None:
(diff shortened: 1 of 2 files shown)


NihalHarish:
The call fn takes a variable number of arguments with different names for different layers.
How do we test that monkey patching doesn't break anything?

vandanavk:
> If call fn takes a variable number of arguments with different names for different layers.
> How do we test that monkey patching doesn't break anything?

TF can take multiple args to call(). For example, dropout layer's call has inputs and training as args. the issue here is that during serialization, TF creates a dictionary mapping of (call_fn_args, args) where call_fn_args is initialized to ["inputs"]. but the name of the input to the layer was called "callable_inputs", so it couldn't find this key in the dictionary mapping and threw an exception. 

possibly a bug in TF, will have to explore more. 

but in the current state, looks like the input argument to call() can only be called "inputs"

PR #345, opened 2020-09-11
Support Inputs and Labels in the dict format

NihalHarish:
### Description of changes:
- flatten dict structure before saving them as tensors.
- Using the same logic used by tf to handle dict structures.

#### Style and formatting:

### DIFF
--- a/smdebug/tensorflow/keras.py
+++ b/smdebug/tensorflow/keras.py
@@ -5,6 +5,7 @@ import functools
 import tensorflow.compat.v1 as tf
 from tensorflow.python.distribute import values
 from tensorflow.python.framework.indexed_slices import IndexedSlices
+from tensorflow.python.util import nest
 
 # First Party
 from smdebug.core.modes import ModeKeys
@@ -497,6 +498,10 @@ class KerasHook(TensorflowBaseHook, tf.keras.callbacks.Callback):
                         else set()
                     )
                 for t_name, t_value in tensors_to_save:
+                    if isinstance(t_value, dict):
+                        # flatten the inputs and labels
+                        # since we cannot convert dicts into numpy
+                        t_value = nest.flatten(t_value)
                     self._save_tensor_to_file(t_name, t_value, collections_to_write)
 
     def _save_metrics(self, batch, logs, force_save=False):
--- a/tests/tensorflow2/test_support_dicts.py
+++ b/tests/tensorflow2/test_support_dicts.py
@@ -36,7 +36,7 @@ def test_support_dicts(out_dir):
     inputs, labels = get_data()
     smdebug_hook = create_hook(out_dir)
     model.fit(inputs, labels, batch_size=16, epochs=10, callbacks=[smdebug_hook])
-
+    model.save(out_dir, save_format="tf")
     trial = create_trial(out_dir)
     assert trial.tensor_names(collection=CollectionKeys.INPUTS) == ["model_input"]
     assert trial.tensor_names(collection=CollectionKeys.OUTPUTS) == ["labels", "predictions"]
--- a/tests/tensorflow2/test_support_dicts.py
+++ b/tests/tensorflow2/test_support_dicts.py
@@ -1,0 +1,42 @@
+# Third Party
+import numpy as np
+import tensorflow as tf
+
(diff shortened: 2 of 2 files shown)


tests/tensorflow2/test_support_dicts.py
```diff
@@ -0,0 +1,42 @@
...
+from smdebug.core.collection import CollectionKeys
+from smdebug.trials import create_trial
+
+
+def get_data():
+    images = np.zeros((64, 224))
+    labels = np.zeros((64, 5))
+    inputs = {"Image_input": images}
+    outputs = {"output-softmax": labels}
+    return inputs, outputs
```
vandanavk, line 16: can we have some test cases for valid and invalid dicts?
empty dicts,
either inputs/output empty (is this possible?)
are the values always valid? - is there a chance they cant be converted to numpy?

is this applicable only to image inputs?

tests/tensorflow2/test_support_dicts.py
```diff
@@ -0,0 +1,42 @@
...
+    return model
+
+
+def test_support_dicts(out_dir):
+    model = create_model()
+    optimizer = tf.keras.optimizers.Adadelta(lr=1.0, rho=0.95, epsilon=None, decay=0.0)
+    model.compile(loss="categorical_crossentropy", optimizer=optimizer)
+    inputs, labels = get_data()
+    smdebug_hook = create_hook(out_dir)
+    model.fit(inputs, labels, batch_size=16, epochs=10, callbacks=[smdebug_hook])
```
vandanavk, line 38: save the model at the end. just to check it works fine

tests/tensorflow2/test_support_dicts.py
vandanavk: are dicts the only other valid format? can inputs be nested lists or something similar?

smdebug/tensorflow/keras.py
```diff
@@ -5,6 +5,7 @@
 import tensorflow.compat.v1 as tf
 from tensorflow.python.distribute import values
 from tensorflow.python.framework.indexed_slices import IndexedSlices
+from tensorflow.python.util import nest
```
vandanavk, line 8: is this valid in both TF1 and TF2?

smdebug/tensorflow/keras.py
```diff
@@ -497,6 +498,10 @@ def save_smdebug_logs(self, logs):
                         else set()
                     )
                 for t_name, t_value in tensors_to_save:
+                    if isinstance(t_value, dict):
```
vandanavk, line 501: can it get to this point in non-eager mode?

tests/tensorflow2/test_support_dicts.py
vandanavk: is this used only with keras fit or can it be used with gradtape too?

smdebug/tensorflow/keras.py
NihalHarish: This import is valid in both TF1 and TF2

smdebug/tensorflow/keras.py
NihalHarish: Both eager and non-eager modes execute this code-block.

tests/tensorflow2/test_support_dicts.py
NihalHarish: The flatten call made by smdebug is only after tf runs the same function on the feed_dict.

Both fit and gradtape execute the same call stack

tests/tensorflow2/test_support_dicts.py
NihalHarish: sure.

tests/tensorflow2/test_support_dicts.py
NihalHarish: > can we have some test cases for valid and invalid dicts?
> empty dicts,
> either inputs/output empty (is this possible?)
> are the values always valid? - is there a chance they cant be converted to numpy?
> 
> is this applicable only to image inputs?

model.fit breaks if the customer passes an empty or invalid feed_dict.

vandanavk (approved):

PR #348, opened 2020-09-11
warn gradtape users  about tf.function support

NihalHarish:
#### Style and formatting:

### DIFF
--- a/smdebug/tensorflow/keras.py
+++ b/smdebug/tensorflow/keras.py
@@ -541,6 +541,9 @@ class KerasHook(TensorflowBaseHook, tf.keras.callbacks.Callback):
             )
             if hasattr(tensor, "numpy"):
                 self._save_tensor_to_file(export_name, tensor.numpy(), input_collection)
+            else:
+                self.logger.warn("cannot save layer values during forward pass with tf.function")
+                return
             # Save Output
             tensor = self.saved_layers[layer_name].layer_output
             export_name = get_export_name_for_keras(layer_name, tensor_type="output", tensor=tensor)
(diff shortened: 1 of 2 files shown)


smdebug/tensorflow/keras.py
```diff
@@ -539,7 +539,8 @@ def _save_layer_input_and_outputs(self):
                 if self._is_collection_being_saved_for_step(CollectionKeys.LAYERS)
                 else set()
             )
-            self._save_tensor_to_file(export_name, tensor.numpy(), input_collection)
+            if hasattr(tensor, "numpy"):
```
vandanavk, line 542: do we have to warn the user if this cannot be saved?

tests/tensorflow2/test_grad_tape_tf_function.py
```diff
@@ -0,0 +1,81 @@
...
+            tf.keras.layers.Flatten(input_shape=(28, 28, 1)),
+            tf.keras.layers.Dense(128, activation="relu"),
+            tf.keras.layers.Dropout(0.2),
+            tf.keras.layers.Dense(10, activation="softmax"),
+        ]
+    )
+    return model
+
+
+def test_gradtape_tf_function(out_dir):
```
vandanavk, line 27: do we have to update the quip doc or general docs about what is supported in this combination?

NihalHarish: Yes. Or we can add this example to the examples repository.

smdebug/tensorflow/keras.py
NihalHarish: Done

vandanavk: will this below log then be called for every tensor that fails?

NihalHarish: The function short circuits after printing the log once for every step.

vandanavk (approved):

PR #349, opened 2020-09-13
list training jobs improvements

Vikas-kum:
### Description of changes:
Earlier list training job would make 50 attempts irrespective. This may be bad because of unnecessary traffic. 
We have changed attempts - 
1) if there are training jobs found with prefix, we break
2) if there are exceptions caught more than 5 times we break. 

#### Style and formatting:


NihalHarish:
> 1. if there are training jobs found with prefix, we break
> 2. if there are exceptions caught more than 5 times we break.

Can we add tests for both these behaviors?

PR #353, opened 2020-09-17
Support Saving Tensors in Graph Mode with add_for_mode

NihalHarish:
### Description of changes:
- This allows users to directly add tensors with `add_for_mode` in `v1.compat` mode

#### Style and formatting:

### DIFF
--- a/smdebug/tensorflow/keras.py
+++ b/smdebug/tensorflow/keras.py
@@ -342,7 +342,6 @@ class KerasHook(TensorflowBaseHook, tf.keras.callbacks.Callback):
                 self._check_and_add_layer_tensor(mode, layer, "weight", w)
 
     def _prepare_non_layer_tensors(self):
-        # for gradients, optimizer_variables
         for coll in self.collection_manager.get_collections().values():
             for tensor_ref in coll.get_tensors():
                 if tensor_ref.name not in self.tensor_to_collections:
--- a/smdebug/tensorflow/keras.py
+++ b/smdebug/tensorflow/keras.py
@@ -352,7 +352,8 @@ class KerasHook(TensorflowBaseHook, tf.keras.callbacks.Callback):
 
     def _prepare_non_layer_tensors(self):
         for coll in self.collection_manager.get_collections().values():
-            for tensor_ref in coll.get_tensors():
+            collection_values = coll.get_tensors()
+            for tensor_ref in collection_values:
                 if tensor_ref.name not in self.tensor_to_collections:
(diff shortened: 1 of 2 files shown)


smdebug/tensorflow/keras.py
```diff
@@ -341,6 +341,14 @@ def _prepare_layers(self, mode):
             for w in weights:
                 self._check_and_add_layer_tensor(mode, layer, "weight", w)
 
+    def _prepare_non_layer_tensors(self):
+        for coll in self.collection_manager.get_collections().values():
+            for tensor_ref in coll.get_tensors():
+                if tensor_ref.name not in self.tensor_to_collections:
+                    self.tensor_to_collections[tensor_ref.name] = {coll}
+                elif coll not in self.tensor_to_collections[tensor_ref.name]:
+                    self.tensor_to_collections[tensor_ref.name].add(coll)
```
vandanavk, line 350: if this is a set, you dont need the``` if coll not in self.tensor_to_collections```.

smdebug/tensorflow/keras.py
```diff
@@ -341,6 +341,14 @@ def _prepare_layers(self, mode):
             for w in weights:
                 self._check_and_add_layer_tensor(mode, layer, "weight", w)
 
+    def _prepare_non_layer_tensors(self):
+        for coll in self.collection_manager.get_collections().values():
+            for tensor_ref in coll.get_tensors():
```
vandanavk, line 346: can you call get_tensors() and save the result before the for loop?

tests/tensorflow2/test_functional_model.py
```diff
@@ -0,0 +1,67 @@
...
+    x = Dropout(0.5)(x)
+    out = Dense(num_classes, activation="softmax")(x)
+
+    model = tf.keras.models.Model(inputs=img_inputs, outputs=out)
+
+    smd_callback = smd.KerasHook(
+        export_tensorboard=False, out_dir=out_dir, include_collections=["custom"]
+    )
+
+    smd_callback.get_collection("custom").add_for_mode([x1], mode=smd.modes.TRAIN)
```
vandanavk, line 53: what is the behavior if this is called with gradtape?

NihalHarish: Gradtape is only supported inside of tf.function or when eager execution is enabled.
This method of saving tensors is only supported when in tf v1 mode.

vandanavk (approved):

PR #357, opened 2020-09-20
Change layer name logic

NihalHarish:
### Description of changes:
- Ensure that the name of the input and output layers do not change for different modes
- Add a test that verifies name strings with a regex pattern

#### Style and formatting:

### DIFF
--- a/smdebug/tensorflow/keras.py
+++ b/smdebug/tensorflow/keras.py
@@ -73,7 +73,6 @@ class KerasHook(TensorflowBaseHook, tf.keras.callbacks.Callback):
         )  # stores tensors custom tensors saved by users every step
         self.saved_layers = dict()
         self.has_registered_model = False
-
         # supports_tf_logs property was introduced in TF 2.3.0
         # it indicates to the framework that the callback is not
         # limited to reading only numpy logs
--- a/smdebug/tensorflow/keras.py
+++ b/smdebug/tensorflow/keras.py
@@ -73,6 +73,7 @@ class KerasHook(TensorflowBaseHook, tf.keras.callbacks.Callback):
         )  # stores tensors custom tensors saved by users every step
         self.saved_layers = dict()
         self.has_registered_model = False
+
         # supports_tf_logs property was introduced in TF 2.3.0
         # it indicates to the framework that the callback is not
         # limited to reading only numpy logs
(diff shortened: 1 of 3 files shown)


tests/tensorflow2/test_keras.py
```diff
@@ -29,6 +30,19 @@
...
+            # WA for TF issue https://github.com/tensorflow/tensorflow/issues/36279
+            tf.keras.layers.Flatten(input_shape=(28, 28, 1)),
+            tf.keras.layers.Dense(128, activation="relu"),
+            tf.keras.layers.Dropout(0.2),
+            tf.keras.layers.Dense(10, activation="softmax"),
+        ]
+    )
+    return model
+
+
```
NihalHarish, line 45: refactored the model def to a fn

tests/tensorflow2/test_keras.py
```diff
@@ -549,6 +545,28 @@ def test_include_regex(out_dir, tf_eager_mode):
...
 
 
+@pytest.mark.slow
+def test_layer_names(out_dir, tf_eager_mode):
+    hook = smd.KerasHook(
+        out_dir,
+        save_config=SaveConfig(save_interval=9),
+        include_collections=[CollectionKeys.LAYERS],
+    )
+    helper_keras_fit(
```
vandanavk, line 555: add a gradtape test too?

smdebug/tensorflow/keras.py
```diff
@@ -757,6 +757,13 @@ def _save_layer_values(self, logs):
         for layer_name, layer_input, layer_output in logs:
             # Cast layer_name to str since it can also be of type bytes
             # when run with mirrored strategy
+            if isinstance(layer_name, tf.Tensor):
```
vandanavk, line 760: isn't this problem there with other tensors that are saved from logs? gradients, metrics etc?

NihalHarish: Gradients already has this logic.
The problem happens from passing strings from the tf layer; so it will not happen with metrics.

vandanavk (approved):

tests/tensorflow2/test_keras.py
```diff
@@ -549,6 +545,44 @@ def test_include_regex(out_dir, tf_eager_mode):
...
+    )
+
+    tr = create_trial_fast_refresh(out_dir)
+    tnames = tr.tensor_names(collection=CollectionKeys.LAYERS)
+    pattern = r"^(flatten|dense|dropout)(_\d+)?\/(inputs|outputs)"
+    for tname in tnames:
+        assert re.match(pattern=pattern, string=tname) is not None
+
+
+@pytest.mark.slow
```
vandanavk, line 570: skip if non eager?

NihalHarish: tf_eager is  not being passed as an argument here.

Issue #363, opened 2020-09-23
Error in atexit


Vikas-kum:
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
File "/usr/local/lib/python3.7/shutil.py", line 566, in move
os.rename(src, real_dst)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/test/events/000000000000/000000000000_worker_0.tfevents.tmp' -> '/tmp/test/events/000000000000/000000000000_worker_0.tfevents'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
File "/codebuild/output/src294/src/s3/00/smdebug/tensorflow/keras.py", line 901, in _cleanup
super()._cleanup()
File "/codebuild/output/src294/src/s3/00/smdebug/core/hook.py", line 528, in _cleanup
self._close_writers()
File "/codebuild/output/src294/src/s3/00/smdebug/tensorflow/base_hook.py", line 362, in _close_writers
self.writer.close()
File "/codebuild/output/src294/src/s3/00/smdebug/core/writer.py", line 229, in close
self.proto_writer.close()
File "/codebuild/output/src294/src/s3/00/smdebug/core/tfevent/event_file_writer.py", line 125, in close
self._ev_writer.close()
File "/codebuild/output/src294/src/s3/00/smdebug/core/tfevent/events_writer.py", line 63, in close
self.tfrecord_writer.close()
File "/codebuild/output/src294/src/s3/00/smdebug/core/tfrecord/record_writer.py", line 81, in close
self._writer.close()
File "/codebuild/output/src294/src/s3/00/smdebug/core/access_layer/file.py", line 53, in close
shutil.move(self.temp_path, self.path)
File "/usr/local/lib/python3.7/shutil.py", line 580, in move
copy_function(src, real_dst)
File "/usr/local/lib/python3.7/shutil.py", line 266, in copy2
copyfile(src, dst, follow_symlinks=follow_symlinks)
File "/usr/local/lib/python3.7/shutil.py", line 120, in copyfile
with open(src, 'rb') as fsrc:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/test/events/000000000000/000000000000_worker_0.tfevents.tmp'
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
File "/usr/local/lib/python3.7/shutil.py", line 566, in move
os.rename(src, real_dst)
[...]

PR #364, opened 2020-09-23
check file exist before moving

Vikas-kum:
### Description of changes:
https://github.com/awslabs/sagemaker-debugger/issues/363

#### Style and formatting:


smdebug/core/access_layer/file.py
```diff
@@ -50,6 +50,10 @@ def close(self):
         """Close the file and move it from /tmp to a permanent directory."""
         self._accessor.close()
         if self.mode in WRITE_MODES:
+            if not os.path.exists(self.temp_path):
+                self.logger.info(
+                    f"Sagemaker-Debugger: Skipping close of file:{self.temp_path} as file doesn't exist"
+                )
             shutil.move(self.temp_path, self.path)
```
NihalHarish, line 57: Shouldn't this be inside an else case?

PR #373, opened 2020-10-14
Pinning the version of tensorflow_datasets package so that it does not require updating TF

leleamol:
### Description of changes:
Installing latest version of tensorflow_datasets (post 4.0.0) requires TF version greater than 2.1.0
Which is causing failure in CI that testing with TF 1.15

#### Style and formatting:


NihalHarish (changes requested):
Can we simply move the installation step from `config/buildspec.yml` to `config/tests.sh`?

If we pin the version of tfds, we will not be able to test later versions which are compatible with TF 2x.

```
if [ "$run_pytest_tensorflow" = "enable" ] ; then
    pip install tensorflow_datasets==3.2.1
    run_for_framework tensorflow
fi

if [ "$run_pytest_tensorflow2" = "enable" ] ; then
    pip install tensorflow_datasets
    run_for_framework tensorflow2
fi
```

NihalHarish (approved):

PR #393, opened 2020-10-30
[WIP]Use get_name in forward hook

NihalHarish:
### Description of bug:
- Users can wrap their modules with helper classes like `DataParallelCriterion` or `DataParallel`
```
custom_loss_module = CustomLossModule()
parallel_custom_loss_module = DataParallelCriterion(custom_loss_module)
```
- The smdebug hook register each module like this:
```
        for name, submodule in module.named_modules():
            assert submodule not in self.module_set, f"Don't register module={module} twice"
            submodule._module_name = name
            self.module_set.add(submodule)
        module._module_name = module._get_name()
        self.module_set.add(module)
```

The problem with the above line is that the `_module_name` attribute is attached to the `parallel_custom_loss_module` and not the nested `custom_loss_module`.

- We would have instead needed to do:
```
submodule.module._module_name = name
```
- When the call to the forward hook is made, the helper module will internally call:
```
    def forward(self, *inputs, **kwargs):
        if not self.device_ids:
            return self.module(*inputs, **kwargs)
```

We should instead simply use:
`module._get_name()` in the `forward_hook` 

#### Style and formatting:

### DIFF
--- a/smdebug/pytorch/hook.py
+++ b/smdebug/pytorch/hook.py
@@ -3,6 +3,7 @@
 # Third Party
 import torch
 import torch.distributed as dist
+from torch.nn.parallel.data_parallel import DataParallel
 
 # First Party
 from smdebug.core.collection import DEFAULT_PYTORCH_COLLECTIONS, CollectionKeys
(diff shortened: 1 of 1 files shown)


Vikas-kum (commented):
Can you check if module.modules will work, as it will give you modules recursively.  https://discuss.pytorch.org/t/module-children-vs-module-modules/4551

PR #405, opened 2020-12-01
Remove logging statement in Keras hook

ndodda-amazon:
### Description of changes:
Removing a logging statement in the Keras hook since it clutters the logs of TF training jobs.

Passing TF 2.3 integration tests: https://tiny.amazon.com/1ddvdftpl/IsenLink

#### Style and formatting:


ndodda-amazon:
@NihalHarish should [this line](https://github.com/awslabs/sagemaker-debugger/pull/405/files#diff-4735c4bf5e5c95768955acc43056504b184783eb3a057db7734a1d93d4ddba0aR577) be removed too?

NihalHarish:
Remove line 577 as well

smdebug/tensorflow/keras.py
```diff
@@ -559,7 +559,6 @@ def _save_layer_input_and_outputs(self):
             )
             t = tensor[0] if isinstance(tensor, list) and len(tensor) else tensor
             if hasattr(t, "numpy") is False:
-                self.logger.warning("cannot save layer values during forward pass with tf.function")
```
NihalHarish, line 562: The same log is present below; we need to remove that as well.

Maybe we can figure out if the condition for this log has been satisfied and then print it outside the loop

NihalHarish (changes requested):

smdebug/tensorflow/keras.py
```diff
@@ -616,10 +618,14 @@ def _save_layer_input_and_outputs(self):
...
+                contain_layer_value = True
+                continue
             else:
                 self._save_tensor_to_file(export_name, tensor, output_collection)
 
+        if contain_layer_value:
```
ndodda-amazon, line 626: This solution mitigates the problem but doesn't really solve the issue. See this example job I ran after building `smdebug` off of this branch: https://tiny.amazon.com/dun1clc1/IsenLink.

There are three possible solutions from here:
1. Remove the logging statement completely, and we can revisit a better way to log the message **(this is my recommendation)**
2. Log this message once, and keep track of a flag to make sure it doesn't get logged again (similar to the profiler config parser)
3. Log the tensor + relevant information with the message so that it isn't duplicate information in each logging message.

JiachengGu (approved):

NihalHarish (approved):

PR #406, opened 2020-12-03
Unblock the debugger CI

NihalHarish:
#### Style and formatting:


ndodda-amazon:
We need to modify [this line](https://github.com/awslabs/sagemaker-debugger/blob/master/config/tests.sh#L53) so that the debugger CI only runs the debugger rules tests.

ndodda-amazon (approved):
LGTM!

PR #408, opened 2020-12-10
Regex match bug

NihalHarish:
### Description of changes:
Bug Description: For the layers and gradients default collection, even if the user specified an include_regex; the regex logic would be ignored and we'd end up capturing all tensors for that default collection.

Changelog:
- Added regex matching logic to the relevant functions to check for regex matches before saving tensors 

#### Style and formatting:


tests/tensorflow2/test_keras.py
```diff
@@ -588,6 +588,33 @@ def test_layer_names(out_dir, tf_eager_mode):
...
+@pytest.mark.slow
+def test_regex_filtering_for_default_collections(out_dir):
+    hook = smd.KerasHook(
+        out_dir,
+        save_config=SaveConfig(save_interval=9),
+        include_collections=[CollectionKeys.LAYERS, CollectionKeys.GRADIENTS],
+    )
+    hook.get_collection("layers").include("^dense")
+    hook.get_collection("gradients").include("gradients/dense")
+    # hook.get_collection(CollectionKeys.LAYERS).include("dense")
```
ndodda-amazon, line 600: Why use
```
hook.get_collection("layers").include("^dense")
hook.get_collection("gradients").include("gradients/dense")
```
instead of
```
hook.get_collection(CollectionKeys.LAYERS).include("^dense")
hook.get_collection(CollectionKeys.GRADIENTS).include("gradients/dense")
```

If this is the expected way for users to specify regex for each collection, then we should be consistent since the hook's `include_collections` parameter seems to accept enum instead of string. Let's have the `get_collection` function accept enum as well.

tests/tensorflow2/test_keras.py
```diff
@@ -588,6 +588,33 @@ def test_layer_names(out_dir, tf_eager_mode):
...
 
 
+@pytest.mark.slow
+def test_regex_filtering_for_default_collections(out_dir):
+    hook = smd.KerasHook(
+        out_dir,
+        save_config=SaveConfig(save_interval=9),
+        include_collections=[CollectionKeys.LAYERS, CollectionKeys.GRADIENTS],
+    )
+    hook.get_collection("layers").include("^dense")
```
ndodda-amazon, line 598: If this is how we specify regex for individual collections, then what does the `include_regex` parameter in the hook do? What happens when `include_regex` is specified with `include_collections`?

tests/tensorflow2/test_keras.py
```diff
@@ -588,6 +588,33 @@ def test_layer_names(out_dir, tf_eager_mode):
...
+        save_config=SaveConfig(save_interval=10),
+        steps=["train"],
+        run_eagerly=True,
+    )
+
+    tr = create_trial_fast_refresh(out_dir)
+    layer_tnames = tr.tensor_names(collection=CollectionKeys.LAYERS)
+    gradient_tnames = tr.tensor_names(collection=CollectionKeys.GRADIENTS)
+    assert len(layer_tnames) == (8 if is_tf_2_2() else 0)
+    assert len(gradient_tnames) == (4 if is_tf_2_2() else 0)
```
ndodda-amazon, line 613: Can we make this check stronger and verify that the captured tensor names actually match the regex that was specified for each of the collections?

smdebug/tensorflow/keras.py
```diff
@@ -600,13 +600,20 @@ def _save_layer_input_and_outputs(self):
...
         for layer_name in self.saved_layers:
             # Save Input
-            tensor = self.saved_layers[layer_name].layer_input
-            export_name = get_export_name_for_keras(layer_name, tensor_type="input", tensor=tensor)
             input_collection = (
                 {self.get_collection(CollectionKeys.LAYERS)}
                 if self._is_collection_being_saved_for_step(CollectionKeys.LAYERS)
                 else set()
             )
+            if bool(input_collection):
```
ndodda-amazon, line 608: Can we make this check better? If `include_collection` is always a set then we can check if the size of the set is nonzero?

ndodda-amazon (changes requested):
Changes look good, left comments about improving testing and general questions.

tests/tensorflow2/test_keras.py
NihalHarish: See the lines below this.
We cannot have hardcoded string values simply because the names of the tensors are assigned numerical prefixes based on when the test is executed.

tests/tensorflow2/test_keras.py
NihalHarish: Customers can do both.

If they use the parameter at the hook level like this:

```
hook = smd.KerasHook(
include_regex=
    )
```

The filter is applied to all collections.
The example in the code, applies the filter to that particular collection only.

ndodda-amazon (approved):
LGTM, feel free to merge once the CI is green. As discussed offline, let's make sure the user experience for setting regex for a single collection is documented somewhere. 

We should also consider making the user experience more simple if users want to specify unique regex for each collection they want to save tensors for. What if `include_collections` could also be a dictionary that mapped user desired collection to corresponding regex for that collection? Not a blocker for this PR but something to consider for the future.

PR #409, opened 2020-12-10
Adding support for pytorch 1.7

leleamol:
### Description of changes:

The code change is required to support PT1.7.

There is change in function signature for torch.autograd.profiler and torch.autograd.profilerConfig in PT 1.7

The integration tests with this change are running: https://tiny.amazon.com/1jxc2ebey/IsenLink
#### Style and formatting:


smdebug/pytorch/hook.py
```diff
@@ -263,9 +263,15 @@ def _collect_torch_profiling_data_if_profiler_enabled(self):
             return
         records = torch.autograd._disable_profiler()
         self.autograd_profiler_enabled = False
-        function_events = torch.autograd.profiler.EventList(
-            torch.autograd.profiler.parse_cpu_trace(records), use_cuda=self.use_cuda
-        )
+        if version.parse(torch.__version__) >= version.parse("1.7"):
```
ndodda-amazon, line 266: Can we move the helper functions `is_pt_1_6`, `is_pt_1_7`, etc from [this test utils file](https://github.com/awslabs/sagemaker-debugger/blob/master/tests/profiler/pytorch/utils.py) to smdebug and use these helper functions throughout this hook file? We have a similar structure in KerasHook There are two benefits to doing this:
- This would reduce code redundancy, so we don't have to repeat this `version.parse` logic everywhere we need to determine the current PT version.
- The util functions define a version more strictly, i.e. `is_pt_1_7` is only `1.7.x`. The difference is that when new PT 1.8.x or future versions are released, this block wouldn't get executed and potentially break (we basically maintain the invariant that unsupported versions = no metrics from autograd and nothing should error).

tests/profiler/pytorch/test_pytorch_profiler.py
```diff
@@ -86,3 +86,5 @@ def test_pytorch_profiler(pytorch_profiler_config_parser, out_dir):
         assert len(events) == 386
     elif is_pt_1_6():
         assert len(events) == 672
+    elif is_pt_1_7():
+        assert 200 <= len(events) <= 555
```
ndodda-amazon, line 90: Does the number of events actually vary from 200 to 555? If so, is there a reason why? This seems like a pretty large gap.

ndodda-amazon (changes requested):
Looks good, left a few comments. Can we also sanity check by running the integration tests for PT 1.6 too?

tests/profiler/pytorch/test_pytorch_profiler.py
leleamol: I was just testing it. Fixed in the code. It should be greater than 220.

leleamol:
Successful PT1.7 integration test : https://tiny.amazon.com/13u1vmw1/IsenLink

leleamol:
Successfult PT1.6 integration test : https://tiny.amazon.com/cphxqyww/IsenLink
The smdebug wheel was installed in sagemaker job using requirements.txt

ndodda-amazon (approved):
Looks good!

PR #411, opened 2020-12-14
Use SMP rank and size when applicable

rahul003:
### Description of changes:
When SMP is used together with Horovod, there'll be multiple horovod 'groups'. Rank and size need to be queried from SMP in such cases. 

This is not a problem for Pytorch as there is a single torch.distributed group there or MXNet as SMP doesn't support MXNet.


smdebug/tensorflow/base_hook.py
```diff
@@ -183,6 +183,17 @@ def _get_worker_name(self) -> str:
...
         self._assert_distribution_strategy()
         if self.distribution_strategy == TFDistributionStrategy.HOROVOD:
+            try:
+                # when model parallel is being used, there will be multiple processes
+                # with same hvd rank, hence use smp.rank
+                import smdistributed.modelparallel.tensorflow as smp
+
+                if smp.core.initialized:
+                    # if smp is in use
+                    return f"worker_{smp.rank()}"
```
Vikas-kum, line 193: Can this block be entered more than once? If yes, I would recommend to  cache this if possible. We have seen calling import like this causes perf penalty. So ensure that the import is not called more than once.

smdebug/tensorflow/base_hook.py
```diff
@@ -260,6 +271,17 @@ def _get_custom_and_default_collections(self) -> Tuple[Set["Collection"], Set["C
     def _get_num_workers(self):
         self._assert_distribution_strategy()
         if self.distribution_strategy == TFDistributionStrategy.HOROVOD:
+            try:
+                # when model parallel is being used, there will be multiple hvd process groups,
+                # hence use smp.size
+                import smdistributed.modelparallel.tensorflow as smp
```
Vikas-kum, line 277: same as above.

smdebug/tensorflow/base_hook.py
```diff
@@ -260,6 +271,17 @@ def _get_custom_and_default_collections(self) -> Tuple[Set["Collection"], Set["C
...
+
+                if smp.core.initialized:
+                    # if smp is in use
+                    return smp.size()
+            except ImportError:
+                pass
+
             import horovod.tensorflow as hvd
```
Vikas-kum, line 285: same applies here also.

rahul003:
ok, will do

smdebug/tensorflow/base_hook.py
rahul003: Modified it

smdebug/core/utils.py
```diff
@@ -317,7 +317,8 @@ def get_tb_worker():
 
 
 def get_distributed_worker():
-    """Get the rank for horovod or torch distributed. If none of them are being used,
+    """This fn should only be used for torch.
```
NihalHarish, line 320: Can we enforce this with something stronger than a comment?
Check if we're running torch for example?

NihalHarish (changes requested):

smdebug/core/utils.py
rahul003: Actually my comment was wrong, but updated this function to work with SMP as well. And made all imports happen only once

NihalHarish (approved):

PR #412, opened 2020-12-15
Updating the version number to 1.0.2

leleamol:
### Description of changes:
Updating the version to 1.0.2 after release of 1.0.1

#### Style and formatting:


smdebug/_version.py
```diff
@@ -1 +1 @@
-__version__ = "1.0.0"
```
NihalHarish, line 1: nit: Why is it not 1.0.1?

leleamol: 1.0.1 is released. Next time when when we create a branch for 1.0.2.  It will automatically inherit this _version.py

NihalHarish: Alright

NihalHarish (approved):

PR #415, opened 2020-12-31
Fix mixed precision import bug for TF 2.4.x

ndodda-amazon:
### Description of changes:
In TensorFlow 2.4, the Keras mixed precision API has moved out of experimental and is now a stable API. As a result, we need to use a different import if the user is using TF 2.4.x

#### Style and formatting:

### DIFF
--- a/smdebug/tensorflow/utils.py
+++ b/smdebug/tensorflow/utils.py
@@ -22,7 +22,10 @@ def does_tf_support_mixed_precision_training():
 
 def supported_tf_variables():
     if does_tf_support_mixed_precision_training():
-        from tensorflow.python.keras.mixed_precision.experimental import autocast_variable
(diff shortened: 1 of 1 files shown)


smdebug/tensorflow/utils.py
```diff
@@ -22,7 +22,10 @@ def does_tf_support_mixed_precision_training():
 
 def supported_tf_variables():
     if does_tf_support_mixed_precision_training():
-        from tensorflow.python.keras.mixed_precision.experimental import autocast_variable
+        if is_tf_version_2_4_x():
```
rahul003, line 25: You might want to make this if version >= 2.4. 
Once an API moves from experimental to mainstream it stays this way in future versions

rahul003 (approved):

PR #421, opened 2021-01-13
Cache TF Versions

NihalHarish:
### Description of changes:
- Simple caching logic to avoid the repeated usage of `version.parse(tf.__version__)`

#### Style and formatting:


connorgoggins (approved):
Non-blocking: might also be worth updating `smdebug/tensorflow/version_check.py`, `tests/tensorflow2/utils.py`, `tests/zero_code_change/smdataparallel_tests/tensorflow2/test_tf2_smdataparallel.py`, `tests/zero_code_change/test_tensorflow2_integration.py`, etc. to use the cached version as well.

ndodda-amazon (changes requested):
Changes look good! Can we cache the PyTorch versions as well for consistency?

ndodda-amazon (approved):

PR #427, opened 2021-01-19
Split inputs to save nested structures

NihalHarish:
### Description of changes:
- Before this PR, inputs, labels and predictions were all saved as a single tensor of type 'list'
- However, tensor lists need to have homogenous objects
- This assumption was proven to be false, so the alternative way to save inputs, labels and predictions is to split them into individual tensors and then retrieve them at the `trial` layer.
- inputs, labels and predictions are now identified with special prefixes: `input_, pred_, label_`
- Since these tensors do not have standardized names defined by tensorflow, it might be confusing to customers to understand the output of smdebug
- So, I have instead added three new functions, `inputs(step=), predictions(step=), and labels(step=)` which will simply collect all the tensors, format them into a python list and return them to the customer.

#### Style and formatting:


smdebug/trials/trial.py
```diff
@@ -364,7 +364,44 @@ def _tensors_in_collection(self, collection) -> set:
             rval.update(self._tensors_matching_regex(regex))
         return rval
 
-    def tensor_names(self, *, step=None, mode=ModeKeys.GLOBAL, regex=None, collection=None) -> list:
+    def inputs(self, step, mode=ModeKeys.GLOBAL):
```
ndodda-amazon, line 367: Where does this function get called? Same with `labels` and `predictions` below.

ndodda-amazon (changes requested):
Can you add a description to this PR summarizing the changes made and what issue(s) it's trying to fix?

smdebug/trials/trial.py
NihalHarish: These are meant to be user facing functions.
I have added tests for them.

smdebug/tensorflow/keras.py
```diff
@@ -538,22 +538,36 @@ def save_gradients_from_logs(self, gradients):
...
+            {self.get_collection(collection_key)}
+            if self._is_collection_being_saved_for_step(collection_key)
+            else set()
+        )
+        if isinstance(tensors_to_save, (dict, list)):
+            tensors_to_save = nest.flatten(tensors_to_save)
+        else:
+            tensors_to_save = [tensors_to_save]
+        idx = 0
+        for t_value in tensors_to_save:
```
ndodda-amazon, line 552: nit: Let's use `enumerate` since it was designed for cases like this.

`for idx, t_value in enumerate(tensors_to_save):`

ndodda-amazon: nit: Let's use `enumerate` since it was designed for cases like this.

`for idx, t_value in enumerate(tensors_to_save):`

No need to define idx or increment it manually.

smdebug/tensorflow/keras.py
```diff
@@ -947,14 +951,35 @@ def _save_layer_values(self, logs):
...
                 # the next layer. Unpacking it speeds up the _make_numpy fn.
                 layer_input = layer_input[0]
-            layer_input_tensor_name = get_export_name_for_keras(str(layer_name), "input")
-            self._save_tensor_to_file(layer_input_tensor_name, layer_input, collections_to_write)
+                self._save_tensor_to_file(
+                    layer_input_tensor_name, layer_input, collections_to_write
+                )
+            else:
+                idx = 0
+                for l_name in layer_input:
```
ndodda-amazon, line 963: nit: Same as above, let's use `enumerate` here

smdebug/tensorflow/keras.py
```diff
@@ -947,14 +951,35 @@ def _save_layer_values(self, logs):
...
+                    layer_input_tensor_name_with_idx = f"{layer_input_tensor_name}_{idx}"
+                    self._save_tensor_to_file(
+                        layer_input_tensor_name_with_idx, l_name, collections_to_write
+                    )
+                    idx += 1
             layer_output_tensor_name = get_export_name_for_keras(str(layer_name), "output")
-            self._save_tensor_to_file(layer_output_tensor_name, layer_output, collections_to_write)
+            if isinstance(layer_output, list):
+                idx = 0
+                for l_output in layer_output:
```
ndodda-amazon, line 972: nit: Same as above, let's use `enumerate` here

smdebug/tensorflow/keras.py
```diff
@@ -947,14 +951,35 @@ def _save_layer_values(self, logs):
                 layer_name = layer_name.name
             elif isinstance(layer_name, bytes):
                 layer_name = str(layer_name, "utf-8")
+            layer_input_tensor_name = get_export_name_for_keras(str(layer_name), "input")
             if len(layer_input) == 1:
```
ndodda-amazon, line 954: nit: Do we still need this if block? Even if there is only one tensor, it might be better to keep the naming schema the same (and still include an index in the name).

smdebug/tensorflow/keras.py
```diff
@@ -947,14 +951,35 @@ def _save_layer_values(self, logs):
...
+                idx = 0
+                for l_name in layer_input:
+                    layer_input_tensor_name_with_idx = f"{layer_input_tensor_name}_{idx}"
+                    self._save_tensor_to_file(
+                        layer_input_tensor_name_with_idx, l_name, collections_to_write
+                    )
+                    idx += 1
             layer_output_tensor_name = get_export_name_for_keras(str(layer_name), "output")
-            self._save_tensor_to_file(layer_output_tensor_name, layer_output, collections_to_write)
+            if isinstance(layer_output, list):
```
ndodda-amazon, line 970: nit: might be cleaner to have a single if check to convert layer_output to a list if it's not already a list.

```
if not isinstance(layer_output, list):
    layer_output = [layer_output]
for l_output in layer_output:
...
```

ndodda-amazon (changes requested):
Changes look good, mostly left comments about nits.

smdebug/tensorflow/keras.py
NihalHarish: There's a performance improvement if we do not process intermediate layer values as lists.
And more often than not, they are single values.

smdebug/tensorflow/keras.py
NihalHarish: Saving layer output values as a list using the _make_numpy fn is slow. Saving them as tensors when they're not lists improves performance

ndodda-amazon (approved):
LGTM! Good work!

PR #431, opened 2021-01-28
Use codebuild env variable to get current branch in profiler integration tests

ndodda-amazon:
### Description of changes:
Unrelated git errors spam the logs of the profiler integration test builds that are run as part of the PR CI. This change fixes that by using a CodeBuild env variable to determine the current branch instead of git commands..

#### Style and formatting:

### DIFF
--- a/config/profiler/run_profiler_integration_tests.sh
+++ b/config/profiler/run_profiler_integration_tests.sh
@@ -5,8 +5,9 @@
 echo "Start profiler integration tests script."
 
 echo "base ref" $CODEBUILD_WEBHOOK_BASE_REF
+echo $(git diff --name-only master $CODEBUILD_WEBHOOK_BASE_REF 2>/dev/null)
 
-CODEBUILD_GIT_BRANCH=$(git branch | grep '^*' | sed 's/* //' )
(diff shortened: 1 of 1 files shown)


config/profiler/run_profiler_integration_tests.sh
```diff
@@ -74,7 +70,6 @@ then
   aws s3 cp s3://smdebug-testing/datasets/cifar-10-python.tar.gz data/cifar-10-batches-py.tar.gz >/dev/null 2>/dev/null # mask output
   aws s3 cp s3://smdebug-testing/datasets/MNIST_pytorch.tar.gz data/MNIST_pytorch.tar.gz >/dev/null 2>/dev/null # mask output
   cd $CODEBUILD_SRC_DIR_TESTS/tests/scripts/pytorch_scripts/data
-  cd data
```
NihalHarish, line 77: Why?

NihalHarish (changes requested):

config/profiler/run_profiler_integration_tests.sh
ndodda-amazon: The line before actually cd's directly into the `data` directory, so this line simply errors. Doesn't affect the behavior of the script but makes the logs more clean.

NihalHarish (approved):

PR #432, opened 2021-02-02
New Xgboost Buildspec

NihalHarish:
### Description of changes:
- Closing this out since the effort was duplicated on #433 

#### Style and formatting:


ndodda-amazon (changes requested):
Changes look good. We need to eventually transition to having one buildspec for each framework + version (to eliminate dependence on env variables set in CodeBuild) and refactoring so that they install packages from a shared requirements.txt to reduce code duplication. Should we handle that in this PR?

Also, is the TF build failure in the CI due to a flaky test?

PR #435, opened 2021-02-03
Remove Spamming Warning Log

NihalHarish:
### Description of changes:
- The warning log should be removed in favor of better documentation of what is supported.

#### Style and formatting:


ndodda-amazon:
If the issue is that this log is being spammed, shouldn't we ensure that this message only gets logged once or a fixed number of times?

NihalHarish:
> If the issue is that this log is being spammed, shouldn't we ensure that this message only gets logged once or a fixed number of times?

Good thought. Let's move the log to a different LOC entirely.

ndodda-amazon (approved):
LGTM! Good work!

PR #443, opened 2021-02-10
Remove Pre-commit from xgboost buildspec

NihalHarish:
### Description of changes:
- codepipeline convert the source to a zip file and hence the source repository loses the `.git` meta 
- running git specific commands causes failures in the nightly

#### Style and formatting:


ndodda-amazon (approved):
As discussed offline, we should eventually have buildspecs for each framework/version (without precommit) that run in CI in nightly. We'll also have a buildspec that runs precommit in CI.

PR #444, opened 2021-02-11
Split inputs to save nested structures

NihalHarish:
### Description of changes:
- Before this PR, inputs, labels and predictions were all saved as a single tensor of type 'list'
- However, tensor lists need to have homogenous objects
- This assumption was proven to be false, so the alternative way to save inputs, labels and predictions is to split them into individual tensors and then retrieve them at the `trial` layer.
- inputs, labels and predictions are now identified with special prefixes: `input_, pred_, label_`
- Since these tensors do not have standardized names defined by tensorflow, it might be confusing to customers to understand the output of smdebug
- So, I have instead added three new functions, `inputs(step=), predictions(step=), and labels(step=)` which will simply collect all the tensors, format them into a python list and return them to the customer.

#### Style and formatting:


ndodda-amazon (commented):
LGTM, will approve after the CI is green.

ndodda-amazon (approved):

PR #449, opened 2021-02-18
Filter Repeating Logs

NihalHarish:
### Description of changes:
- Filter out logs that repeat more than 5 times to prevent log spam

#### Style and formatting:

### DIFF
--- a/smdebug/core/logger.py
+++ b/smdebug/core/logger.py
@@ -5,6 +5,9 @@ import socket
 import sys
 from collections import defaultdict
 
+# First Party
+from smdebug.core.config_constants import LOG_DUPLICATION_THRESHOLD
+
 _logger_initialized = False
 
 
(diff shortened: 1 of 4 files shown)


leleamol (commented):
The fix looks good. In addition, can we reduce the logging level to debug for 
logger.info(
                    "Checkpoints not updated. There are no checkpoint files created yet, to be updated"
                )

NihalHarish:
> Checkpoints not updated. There are no checkpoint files created yet, to be updated

Do I change the level for all the logs in the file or just the one you have mentioned?

leleamol (approved):

smdebug/core/logger.py
```diff
@@ -19,6 +20,18 @@ def filter(self, record):
         return record.levelno < self.level
 
 
+class DuplicateLogFilter:
+    """Filters duplicate messages to prevent spamming users"""
+
+    def __init__(self):
+        self.msgs = defaultdict(int)
+        self.repeat_threshold = 5
```
ndodda-amazon, line 28: Can we lower this to 2 or 3?

Also, can we make this a constant?

ndodda-amazon (approved):
LGTM! Left a minor comment.

PR #425, opened 2021-01-16
Redo of PR 411 Use Smp rank and size when applicable

rahul003:
### Description of changes:
Redo of https://github.com/awslabs/sagemaker-debugger/pull/411 

Fixed the issue of smdataparallel being imported globally by moving it within the check_dataparallel_env function and caching the imported library

#### Style and formatting:

I have run `pre-commit install` to ensure that auto-formatting happens with every commit.

#### Issue number, if available

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #451, opened 2021-02-23
SMDDP should use size() and rank() for TF jobs

ndodda-amazon:
### Description of changes:
Fixing a bug introduced by #425 where `get_size()` and `get_rank()` were used for all jobs, when the correct API for TF jobs is actually is actually `size()` and `rank()`.

This bug was missed before because we do not have any tests that run TF SMDDP jobs.

#### Style and formatting:

### DIFF
--- a/config/buildspec.yml
+++ b/config/buildspec.yml
@@ -30,7 +30,7 @@ phases:
         - sudo apt-get install unzip -qq -o=Dpkg::Use-Pty=0
         - cd $CODEBUILD_SRC_DIR  && chmod +x config/protoc_downloader.sh && ./config/protoc_downloader.sh
         - pip install --upgrade pip==20.3.3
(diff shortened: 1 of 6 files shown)


ChaiBapchya (commented):
I haven't tested it locally on my end. But the API spec looks right now
TF uses `size()` 
```
 smdistributed.dataparallel.tensorflow.size()
```

PT uses `get_world_size()`
```
smdistributed.dataparallel.torch.distributed.get_world_size()
```

[1] https://sagemaker.readthedocs.io/en/stable/api/training/sdp_versions/v1.0.0/smd_data_parallel_tensorflow.html#smdistributed.dataparallel.tensorflow.size
[2] https://sagemaker.readthedocs.io/en/stable/api/training/sdp_versions/v1.0.0/smd_data_parallel_pytorch.html#smdistributed.dataparallel.torch.distributed.get_world_size

leleamol (approved):

ChoiByungWook:
User is reporting a similar issue: https://github.com/aws/sagemaker-python-sdk/issues/2170

PR #454, opened 2021-03-01
bugfix for timelinewriter

NRauschmayr:
### Description of changes:
Timelinewriter has a bug where during file-rotation the file would be overwritten with the same filename and such events are overwritten.

 This bug only happens when the amount of data emitted is high for instance during detailed profiling. Here is an example of the merged timeline without the bugfix:

Screen Shot 2021-03-01 at 1 34 36 PM

And with the bugfix:
Screen Shot 2021-03-01 at 1 33 48 PM

The root cause is that we overwrite the filename with the end timestamp of an event instead of the begin timestamp: some events such as step duration take a very long time and with detailed profiling a lot of other events are emitted in the meantime, leading to the file rotation being triggered because of max file size.
#### Style and formatting:


leleamol (approved):
LGTM

PR #460, opened 2021-03-09
bugfix for timelinewriter

NRauschmayr:
### Description of changes:
more details in https://github.com/awslabs/sagemaker-debugger/pull/454

#### Style and formatting:


leleamol (approved):
Approving the change but can we verify that the hourly directories are getting created event if rotation_policy.file_close_interval != 60

smdebug/core/tfevent/timeline_file_writer.py
```diff
@@ -306,13 +306,15 @@ def _should_rotate_now(self, now_in_us):
...
-            return True
+        #        if diff_in_hours != 0:
+        #            return True
 
         if diff_in_seconds > rotation_policy.file_close_interval:
             return True
 
         if file_size > rotation_policy.file_max_size:
+            if now_in_us != self.last_event_end_time_in_us:
+                self.last_event_end_time_in_us = now_in_us
```
NihalHarish, line 317: Can we add a test for these changes?

NihalHarish (changes requested):

smdebug/core/tfevent/timeline_file_writer.py
```diff
@@ -306,13 +306,15 @@ def _should_rotate_now(self, now_in_us):
         file_size, diff_in_seconds, diff_in_hours = self._get_rotation_info(now_in_us)
         rotation_policy = self._profiler_config_parser.config.trace_file.rotation_policy
 
-        if diff_in_hours != 0:
-            return True
+        #        if diff_in_hours != 0:
+        #            return True
```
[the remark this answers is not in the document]
NRauschmayr: i removed it

smdebug/core/tfevent/timeline_file_writer.py
NRauschmayr: This line fixes the problem of overwriting the same filename where we basically lost data during detailed profiling. I have tested this with long running training jobs and here is the result that shows that without the fix merged timeline is missing events:
image


With bugfix:
image

The new testcases that will be added to the CI ensure this part is tested.

NihalHarish (approved):

PR #462, opened 2021-03-14
Autoconfigure Losses Collection For  Xgboost

NihalHarish:
### Description of changes:
- All the deep learning frameworks (mxnet, tensorflow, pytorch) have a collection called "losses", which saves loss tensors.
- xgboost calls the same tensor "metrics" and the collection "losses" has no meaning.
- The LossNotDecreasing Rule is a builtin rule, that by default attempts to save the "losses" collection.
- This leads to the error `InvalidCollection` being raised, which is meant to warn users about incorrect collection configurations, especially when they create custom collections.
- However, since the LossNotDecreasing Rule is 1P rule and must not exhibit this behavior.
- With this PR, I simply add the Losses collection with the same regex as Metrics so that unsuspecting customers do not run into issues with the 1P rule and the LossNotDecreasing Rule will work as expected.
- Documentation of the use of the LossNotDecreasing rule for xgboost overrides losses with metrics; see [link](https://github.com/aws/amazon-sagemaker-examples/blob/master/sagemaker-debugger/xgboost_builtin_rules/xgboost-regression-debugger-rules.ipynb), however it is not made explicit and I believe that users should not have to configure one particular builtin rule for a certain framework
- **Note:** This change only comes into effect if the user configures a losses collection without a regex or use the LossNotDecreasing 1P Rule

#### Style and formatting:

### DIFF
--- a/smdebug/xgboost/collection.py
+++ b/smdebug/xgboost/collection.py
@@ -14,7 +14,8 @@ class CollectionManager(BaseCollectionManager):
             self.create_collection(c)
         self.get(CollectionKeys.HYPERPARAMETERS).include("^hyperparameters/.*$")
         self.get(CollectionKeys.METRICS).include("^[a-zA-z]+-[a-zA-z0-9]+$")
-        self.get(CollectionKeys.LOSSES).include("^[a-zA-z]+-[a-zA-z0-9]+$")
+        # we exclude map and auc since they're not loss metrics
+        self.get(CollectionKeys.LOSSES).include("^[a-zA-z]+-.+(?<!(auc|map))$")
         self.get(CollectionKeys.PREDICTIONS).include("^predictions$")
         self.get(CollectionKeys.LABELS).include("^labels$")
         self.get(CollectionKeys.FEATURE_IMPORTANCE).include("^feature_importance/.*")
--- a/smdebug/xgboost/collection.py
+++ b/smdebug/xgboost/collection.py
@@ -14,6 +14,7 @@ class CollectionManager(BaseCollectionManager):
             self.create_collection(c)
         self.get(CollectionKeys.HYPERPARAMETERS).include("^hyperparameters/.*$")
         self.get(CollectionKeys.METRICS).include("^[a-zA-z]+-[a-zA-z0-9]+$")
+        self.get(CollectionKeys.LOSSES).include("^[a-zA-z]+-[a-zA-z0-9]+$")
         self.get(CollectionKeys.PREDICTIONS).include("^predictions$")
         self.get(CollectionKeys.LABELS).include("^labels$")
         self.get(CollectionKeys.FEATURE_IMPORTANCE).include("^feature_importance/.*")
--- a/smdebug/xgboost/hook.py
+++ b/smdebug/xgboost/hook.py
@@ -1,5 +1,6 @@
 # Standard Library
 import os
+import re
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 # Third Party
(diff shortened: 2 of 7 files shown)


smdebug/xgboost/collection.py
```diff
@@ -14,6 +14,7 @@ def _register_default_collections(self):
             self.create_collection(c)
         self.get(CollectionKeys.HYPERPARAMETERS).include("^hyperparameters/.*$")
         self.get(CollectionKeys.METRICS).include("^[a-zA-z]+-[a-zA-z0-9]+$")
+        self.get(CollectionKeys.LOSSES).include("^[a-zA-z]+-[a-zA-z0-9]+$")
```
NRauschmayr, line 17: I just have a minor comment: if we make the loss collection the same like metrics collection, then it can include metrics such as error, rmse, mae, mape (which is expected to decrease during training) and metrics such as aoc, map (which is expected to increase during training). So the loss not decreasing rule could potentially create false positives.

NihalHarish: The name of the metric can change across training jobs, right?
Maybe we should update this based on hyperparameter?

smdebug/xgboost/hook.py
```diff
@@ -18,7 +18,13 @@
 
 # Local
 from .collection import CollectionManager
-from .utils import get_content_type, get_dmatrix, parse_tree_model, validate_data_file_path
+from .utils import (
+    _filter_increasing_metric_names,
```
NihalHarish, line 22: Added this `_filter_increasing_metric_names` to the list of imports.

smdebug/xgboost/utils.py
```diff
@@ -128,6 +129,13 @@ def _get_num_valid_libsvm_features(libsvm_line):
         return 0
 
 
+def _filter_increasing_metric_names(metrics: List[Tuple]):
+    # auc and map are metrics that increase in value
+    # and should not be classified as losses
+    p = re.compile("^[a-zA-z]+-(auc|map)$")
+    return [metric_name for metric_name, metric_value in metrics if p.match(metric_name)]
```
ndodda-amazon, line 136: nit: s/metric_value/_ because it's not being used in the list comprehension

smdebug/xgboost/hook.py
```diff
@@ -219,6 +228,18 @@ def write_hyperparameters(self, env: CallbackEnv):
...
 
+    def write_losses(self, env: CallbackEnv):
+        # Get loss metric at the current boosting round
+        # loss metrics will have already been saved by write_metrics if the metrics collection is included.
+        if self._is_collection_being_saved_for_step(CollectionKeys.METRICS):
+            return
+        else:
+            evaluation_metrics = env.evaluation_result_list
+            increasing_metric_names = _filter_increasing_metric_names(evaluation_metrics)
+            for metric_name, metric_data in evaluation_metrics:
```
ndodda-amazon, line 239: This could be slightly refactored to be more simple. 

```
evaluation_metrics = env.evaluation_result_list
p = re.compile("^[a-zA-z]+-(auc|map)$") # regex for matching increasing metrics
for metric_name, metric_data in evaluation_metrics:
    if p.match(metric_name):
        continue
    self._save_for_tensor(metric_name, metric_data)
    
```
Thoughts?

ndodda-amazon (changes requested):
Changes look good. I have a few questions though.
- This change should allow the LossNotDecreasing rule to work with xgboost right?
- If so, can you test that the LossNotDecreasing rule works with xgboost given this change?

NRauschmayr (approved):

NihalHarish:
@ndodda-amazon : I tested the changes with this [notebook](https://delete-if-found.notebook.us-east-1.sagemaker.aws/notebooks/test-loss-not-decreasing-rule.ipynb ).

Take a look

ndodda-amazon (approved):
Thanks @NihalHarish, the notebook you linked looks good to me. Let's make sure to follow up with adding a sagemaker test with xgboost and lossnotdecreasing.

Otherwise, LGTM, good work! Feel free to merge after the CI passes.

PR #463, opened 2021-03-16
[WIP] Use file metadata to determine whether profiler config should be reloaded.

ndodda-amazon:
### Description of changes:
For each step, we need to determine if the profiler config JSON has changed, and if so, we should reload the profiler config. Currently, we reload the JSON into memory and physically check whether the file contents have changed in order to determine if the profiler config should be reloaded. However, this may pose problems for performance at scale because we would be loading a JSON object into memory at each step.

This change replaces the above check by inspecting the file metadata for the last modified time. If the last modified time has changed, that means the file has changed and we should reload the profiler config. This is done without loading the JSON into memory (see tests, which verify that the config file is not accessed (read into memory) if the file has not been modified). 

#### Style and formatting:

### DIFF
--- a/smdebug/profiler/utils.py
+++ b/smdebug/profiler/utils.py
@@ -306,5 +306,5 @@ def get_last_modified_time(filepath):
     """
     Get the last time that the file at the given filepath was modified, in the form of a datetime object.
     """
-    last_modified_time = Path(filepath).stat().st_mtime
+    last_modified_time = os.path.getmtime(filepath)
     return datetime.fromtimestamp(last_modified_time)  # get the last time the config was modified
--- a/smdebug/profiler/utils.py
+++ b/smdebug/profiler/utils.py
@@ -300,3 +300,11 @@ def stop_smdataparallel_profiler(smdataparallel, base_dir):
         ensure_dir(new_file_name)
         if os.path.exists(smdataparallel_temp_file):
             shutil.move(smdataparallel_temp_file, new_file_name)
+
+
+def get_last_modified_time(filepath):
+    """
+    Get the last time that the file at the given filepath was modified, in the form of a datetime object.
+    """
+    last_modified_time = Path(filepath).stat().st_mtime
+    return datetime.fromtimestamp(last_modified_time)  # get the last time the config was modified
--- a/tests/profiler/core/test_profiler_config_parser.py
+++ b/tests/profiler/core/test_profiler_config_parser.py
@@ -409,26 +409,26 @@ def test_update_step_profiler_config_parser(
 
     profiler_config_parser.load_config()
 
-    # check that reloading the config when it hasn't changed won't change the config fields.
-    assert profiler_config_parser.config.detailed_profiling_config.start_step == 5
-    assert profiler_config_parser.config.detailed_profiling_config.num_steps == 2
-
     # check that the config wasn't loaded into memory again.
     last_accessed_time = get_last_accessed_time(step_profiler_config_parser_path)
     assert first_accessed_time == last_accessed_time
 
+    # check that reloading the config when it hasn't changed won't change the config fields.
+    assert profiler_config_parser.config.detailed_profiling_config.start_step == 5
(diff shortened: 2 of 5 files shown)


tests/profiler/core/test_profiler_config_parser.py
```diff
@@ -368,17 +380,23 @@ def test_case_insensitive_profiler_config_parser(case_insensitive_profiler_confi
 
 
 def test_update_step_profiler_config_parser(
```
NihalHarish, line 382: Can you run this test inside of a loop, to ensure that we do not have flaky behavior.

smdebug/profiler/utils.py
```diff
@@ -300,3 +300,11 @@ def stop_smdataparallel_profiler(smdataparallel, base_dir):
         ensure_dir(new_file_name)
         if os.path.exists(smdataparallel_temp_file):
             shutil.move(smdataparallel_temp_file, new_file_name)
+
+
+def get_last_modified_time(filepath):
+    """
+    Get the last time that the file at the given filepath was modified, in the form of a datetime object.
+    """
+    last_modified_time = Path(filepath).stat().st_mtime
```
NihalHarish, line 309: Can you explain in comments, the reasons for the choice of this implementation?
A simple link to documentation might also suffice.
https://docs.python.org/3/library/stat.html#stat.ST_MTIME

NihalHarish (changes requested):

tests/profiler/core/test_profiler_config_parser.py
ndodda-amazon: Would be better to trivially use `pytest.mark.parametrize` so that the test itself doesn't change, but we simply run it multiple times and ensure it passes each time.

How many times do you suggest the test be run? 3?

ndodda-amazon: Also, first and foremost, can you explain why this behavior might be flaky? The workflow seems to be clear cut:
1. Set up the profiler config JSON
2. Set up the profiler config parser object (which automatically loads the config)
3. Load the config again
4. Verify the last modified time has not changed
5. Replace the config JSON
6. Load the config again
7. Verify the last modified time has changed

tests/profiler/core/test_profiler_config_parser.py
```diff
@@ -389,32 +407,48 @@ def test_update_step_profiler_config_parser(
...
+    last_accessed_time = get_last_accessed_time(step_profiler_config_parser_path)
+    assert first_accessed_time == last_accessed_time
+
+    # check that reloading the config when it hasn't changed won't change the config fields.
     assert profiler_config_parser.config.detailed_profiling_config.start_step == 5
     assert profiler_config_parser.config.detailed_profiling_config.num_steps == 2
 
-    # check that reloading the config when it has changed will update the config fields.
-    monkeypatch.setenv("SMPROFILER_CONFIG_PATH", new_step_profiler_config_parser_path)
+    shutil.copy(new_step_profiler_config_parser_path, step_profiler_config_parser_path)
```
ndodda-amazon, line 420: Tests are currently failing since this call has different behavior for Linux and Unix (for Linux, this copies the file metadata too).

Will need to fix this by finding a better API for copying files or writing the JSON manually.

smdebug/profiler/utils.py
ndodda-amazon: The advantage of relying on file metadata to determine whether a profiler config should be reloaded is stated in the PR description - we cut down on the amount of times the JSON needs to be loaded into memory.

The choice of specifically using `pathlib` to get the file metadata was arbitrary - we can just as easily use `os` via `os.path.getmtime(path)`. In fact, I probably will switch to `os.path.getmtime(path)` since it appears cleaner.

ndodda-amazon:
Closing for now since the root cause of the CI failures has to do with Linux flakiness in the metadata of the profiler config JSON not being updated. Will reopen once I've fixed this issue.

PR #464, opened 2021-03-17
Use file metadata to determine whether profiler config should be reloaded.

ndodda-amazon:
### Description of changes:
(Unable to reopen #463 so I'm creating a new PR).

For each step, we need to determine if the profiler config JSON has changed, and if so, we should reload the profiler config. Currently, we reload the JSON into memory and physically check whether the file contents have changed in order to determine if the profiler config should be reloaded. However, this may pose problems for performance at scale because we would be loading a JSON object into memory at each step.

This change replaces the above check by inspecting the file metadata for the last modified time. If the last modified time has changed, that means the file has changed and we should reload the profiler config. This is done without loading the JSON into memory (see tests, which verify that the config file is not accessed (read into memory) if the file has not been modified).

#### Style and formatting:

### DIFF
--- a/smdebug/profiler/utils.py
+++ b/smdebug/profiler/utils.py
@@ -306,5 +306,5 @@ def get_last_modified_time(filepath):
     """
     Get the last time that the file at the given filepath was modified, in the form of a datetime object.
     """
-    last_modified_time = Path(filepath).stat().st_mtime
+    last_modified_time = os.path.getmtime(filepath)
     return datetime.fromtimestamp(last_modified_time)  # get the last time the config was modified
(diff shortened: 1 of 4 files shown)


smdebug/profiler/utils.py
```diff
@@ -300,3 +300,11 @@ def stop_smdataparallel_profiler(smdataparallel, base_dir):
...
         if os.path.exists(smdataparallel_temp_file):
             shutil.move(smdataparallel_temp_file, new_file_name)
+
+
+def get_last_modified_time(filepath):
+    """
+    Get the last time that the file at the given filepath was modified, in the form of a datetime object.
+    """
+    last_modified_time = os.path.getmtime(filepath)
+    return datetime.fromtimestamp(last_modified_time)  # get the last time the config was modified
```
NihalHarish, line 310: I want to make sure this does not return false positives.

What if an agent owned by the platform team touches the file which causes changes in the `last_modified_time` value?

Checking for a change in file_size itself might provide a stronger signal.
What do you think?

NihalHarish: Interesting. I see you've used getatime below and getmtime here.

What is the difference between the two?

NihalHarish: Are there any long term risks? Do we have to worry about the OS the container is running on?

PR #466, opened 2021-03-18
[bugfix] wrap read metadata.json in a try except block

NihalHarish:
### Description of changes:
- if the `metadata.json` file is empty there is potential that the hook will crash.

#### Style and formatting:

### DIFF
--- a/smdebug/core/state_store.py
+++ b/smdebug/core/state_store.py
@@ -83,7 +83,11 @@ class StateStore:
         """
         if os.path.exists(self._states_file):
             with open(self._states_file) as json_data:
-                parameters = json.load(json_data)
+                try:
+                    parameters = json.load(json_data)
+                except json.decoder.JSONDecodeError:
+                    logger.warning(f"{self._states_file} was empty")
+                    return
             for param in parameters:
                 ts_state = dict()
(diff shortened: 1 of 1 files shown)


smdebug/core/state_store.py
```diff
@@ -83,7 +83,11 @@ def _read_states_file(self):
         """
         if os.path.exists(self._states_file):
             with open(self._states_file) as json_data:
-                parameters = json.load(json_data)
+                try:
+                    parameters = json.load(json_data)
+                except json.decoder.JSONDecodeError:
+                    logger.warning(f"{self._states_file} was empty")
```
NihalHarish, line 89: This should not have potential to create a log flood since the str is constant.
Keeping the `level=warning` since we'd ideally want feedback if this behavior is seen by the customers too often

NihalHarish: Open to opinions.

ndodda-amazon: In what case would the states file be empty? If it's because of a user misconfiguration of some sort, then yes I agree with making this a warning message. Otherwise, I'm not sure. What are the repercussions for the user if the states file is empty?

PR #470, opened 2021-03-23
[Do-Not-Merge] Ignore sagemaker-uploading

NihalHarish:
### Description of changes:
- We should ignore special status files that can be added by Sagemaker before determining if we should update the checkpointing directory.

#### Style and formatting:

### DIFF
--- a/smdebug/_version.py
+++ b/smdebug/_version.py
@@ -1,1 +1,1 @@
-__version__ = "1.0.8check_fix"
+__version__ = "1.0.8b"
--- a/smdebug/_version.py
+++ b/smdebug/_version.py
@@ -1,1 +1,1 @@
-__version__ = "1.0.8"
+__version__ = "1.0.8check_fix"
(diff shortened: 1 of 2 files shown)


smdebug/_version.py
```diff
@@ -1 +1 @@
-__version__ = "1.0.8"
+__version__ = "1.0.8b"
```
ndodda-amazon, line 1: Why?

smdebug/core/state_store.py
```diff
@@ -33,6 +33,7 @@ def _get_checkpoint_files_in_dir(self, cp_dir):
                     file != METADATA_FILENAME
                     and file != METADATA_FILENAME_S3_UPLOADED
                     and "sagemaker-uploaded" not in file
+                    and "sagemaker-uploading" not in file
```
ndodda-amazon, line 36: nit: This won't scale well. Can we keep track of a list of files not to append and then do something like:

```
if all([metadata_file not in file for metadata_file in metadata_files]) 
```

PR #471, opened 2021-03-23
Add Debug Logs

NihalHarish:
#### Style and formatting:

### DIFF
--- a/smdebug/_version.py
+++ b/smdebug/_version.py
@@ -1,1 +1,1 @@
-__version__ = "1.0.8"
+__version__ = "1.0.8logging"
--- a/smdebug/tensorflow/keras.py
+++ b/smdebug/tensorflow/keras.py
@@ -166,6 +166,7 @@ class KerasHook(TensorflowBaseHook, tf.keras.callbacks.Callback):
         # This function is called by the hook in the AWS TF codebase
         # It attaches a hook to every layer of the model to capture
         # layer values
+        self.logger.debug("Smdebug Register Model")
         self.model = model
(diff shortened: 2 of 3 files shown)


smdebug/_version.py
```diff
@@ -1 +1 @@
-__version__ = "1.0.8"
+__version__ = "1.0.8c0"
```
ndodda-amazon, line 1: Why?

smdebug/core/writer.py
```diff
@@ -155,6 +156,7 @@ def write_tensor(
         tag = tname
         tensor_proto = make_tensor_proto(nparray_data=value, tag=tag)
         s = Summary(value=[Summary.Value(tag=tag, metadata=smd, tensor=tensor_proto)])
+        self.logger.debug(f"smdebug: writing tensor: {tname}")
```
ndodda-amazon, line 159: Shouldn't the logger itself handle prepending smdebug as an identifier to each logging message?

smdebug/tensorflow/keras.py
```diff
@@ -725,6 +727,7 @@ def _remove_fetches_and_callbacks(self, mode):
 
     def _prepare_collections_for_tf2(self):
         self._prepare_collections()
+        self.logger.debug("SMDEBUG: Preparing Collections")
```
ndodda-amazon, line 730: Case of "smdebug" is inconsistent. There should be some higher level wrapper that manages prepending "smdebug" so that it isn't expected for the writer of the log to remember to prepend "smdebug".

Issue #418, opened 2021-01-09
TypeError: get() takes no keyword arguments - breaks training jobs


robwhelan:
I have been fine-tuning `distilbert` from the HuggingFace Transformers project. When calling `trainer.train()`, somewhere `smdebug` tries to call `os.environ.get()` and I get the above error.

There are no other messages.

It affects this line: `/smdebug/core/logger.py", line 51, in get_logger` whether or not I set `debugger_hook_config=False`

```python
from datasets import load_dataset
from transformers import AutoModelForCausalLM
from transformers import AutoTokenizer
from transformers import Trainer, TrainingArguments
import argparse
import json
import logging
import torch
import torch.distributed as dist
import torch.nn as nn
import torch.nn.functional as F
import torch.optim as optim
import torch.utils.data
import torch.utils.data.distributed
import math
import os
import sys
from dataclasses import dataclass, field
from typing import Optional

#logger = logging.getLogger(__name__)
#logger.setLevel(logging.DEBUG)
#logger.addHandler(logging.StreamHandler(sys.stdout))

#logging.basicConfig(
#        format="%(asctime)s - %(levelname)s - %(name)s -   %(message)s",
#        datefmt="%m/%d/%Y %H:%M:%S",
#        level=logging.WARN,
#    )

def group_texts(examples):
    # Concatenate all texts.
    block_size = 64
    concatenated_examples = {k: sum(examples[k], []) for k in examples.keys()}
    total_length = len(concatenated_examples[list(examples.keys())[0]])
    # We drop the small remainder, we could add padding if the model supported it instead of this drop, you can
        # customize this part to your needs.
    total_length = (total_length // block_size) * block_size
    # Split by chunks of max_len.
    result = {
        k: [t[i : i + block_size] for i in range(0, total_length, block_size)]
        for k, t in concatenated_examples.items()
    }
    result["labels"] = result["input_ids"].copy()
    return result


def train(args):
    train_file = os.path.join(args.train_dir, "new_train.txt")
    valid_file = os.path.join(args.valid_dir, "new_valid.txt")

    model_checkpoint = "distilgpt2"
    
    #datasets = load_dataset('text', data_files={'train': train_file, 'test': valid_file, 'valid': valid_file})
    datasets = load_dataset('wikitext', 'wikitext-2-raw-v1')
    tokenizer = AutoTokenizer.from_pretrained(model_checkpoint, use_fast=True)
    tokenizer.pad_token = tokenizer.eos_token
[...]

PR #473, opened 2021-03-24
[DO Not Merge] Avoid using 'default' keyword while invoking 'get' method

leleamol:
### Description of changes:
As mentioned in the issue #418 customers may see the error because of having 'default' keyword in the 'get()' call.

This change avoid using this keyword argument in 'get()' call to avoid potential issues.

As mentioned in the issue, the tokenizer in transformers package is modifying the data type of os.environ from os.Environ to 'dict'.  The change in this PR is for protecting smdebug from such changes while keeping the behavior same.  

#### Style and formatting:


NihalHarish:
Let's not merge this.
If this is the correct way to handle env variables then be it.

PR #481, opened 2021-04-08
Updated heatmap visualizations

NRauschmayr:
### Description of changes:
I updated the heatmap to provide better visualizations for large scale distributed training jobs for instance users can now aggregate metrics per worker node.
Before aggregation per worker node was not supported, so the visualization becomes very overwhelming see the following image which shows utilization per GPU and CPU core
image
With the updated heatmap, users can aggregate utilization per worker node like in the following image:  
Screen Shot 2021-04-13 at 12 24 43 PM

#### Style and formatting:

### DIFF
--- a/smdebug/profiler/analysis/notebook_utils/heatmap.py
+++ b/smdebug/profiler/analysis/notebook_utils/heatmap.py
@@ -1,16 +1,15 @@
 # Standard Library
 import re
+from copy import deepcopy
 
 # Third Party
 import bokeh
 import numpy as np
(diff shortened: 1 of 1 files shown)


NihalHarish (changes requested):
Can we add tests for this?

NihalHarish:
We need to add a way to test visualizations going forward; for now I will approve these changes.

NihalHarish (approved):

atqy:
Just want to ask, why is the `update_data` function deleted?

PR #483, opened 2021-04-23
Implement error handling for TF2

ndodda-amazon:
### Description of changes:
Error handler to catch all errors originating from smdebug and its dependencies. This is instantiated as a global util object and wrapped around smdebug functions.
    
Currently, the error handler is designed to catch all errors that could come up for the default smdebug configuration, or in other words, the default debugger and profiler configuration. The error handler is wrapped around all such externally facing smdebug functions (i.e. called from the ML framework). smdebug functions that would not get called with the default smdebug configuration (i.e. called by user manually) are not in scope and would not get wrapped by the error handler.
    
If an error handler catches an error, smdebug functionality is disabled for the rest of training.
#### Style and formatting:


smdebug/core/hook.py
```diff
@@ -395,6 +398,12 @@ def has_default_hook_configuration(self):
             return True
         return False
 
+    def has_default_profiler_configuration(self):
```
connorgoggins, line 401: nit: if this function is only used internally (in `has_default_configuration()`), maybe prepend an underscore to the function name? Otherwise, an error handler decorator should be added to this function.

tests/tensorflow2/test_error_handling.py
```diff
@@ -0,0 +1,211 @@
...
+    profiler_config_path,
+    out_dir,
+    hook_class_with_keras_callback_error,
+    custom_configuration_error_message,
+    stack_trace_filepath,
+):
+    """
+    Test that the error handler does not catch errors when a custom smdebug configuration of smdebug is used.
+
+    This hook needs to be initialized during its corresponding test, because the error handler is configured to a hook
```
ndodda-amazon, line 187: nit: s/This/Each

tests/tensorflow2/test_error_handling.py
```diff
@@ -0,0 +1,211 @@
...
+    return HookWithBadKerasAndLayerCallback
+
+
+@pytest.fixture
+def profiler_config_path(config_folder):
+    return os.path.join(config_folder, "test_tf2_profiler_config_parser_by_time.json")
+
+
+@pytest.fixture(autouse=True)
+def set_up(out_dir, stack_trace_filepath):
```
ndodda-amazon, line 99: nit: Add docstring

smdebug/tensorflow/keras.py
```diff
@@ -1059,12 +1074,15 @@ def _on_any_batch_end(self, batch, mode, logs=None):
...
     def on_train_batch_end(self, batch, logs=None):
         self._on_any_batch_end(batch, ModeKeys.TRAIN, logs=logs)
 
+    @error_handler.catch_smdebug_errors()
     def on_test_batch_end(self, batch, logs=None):
         self._on_any_batch_end(batch, ModeKeys.EVAL, logs=logs)
 
+    @error_handler.catch_smdebug_errors()
     def on_predict_batch_end(self, batch, logs=None):
         self._on_any_batch_end(batch, ModeKeys.PREDICT, logs=logs)
```
connorgoggins, line 1088: Why isn't a decorator necessary for `wrap_optimizer()`?

tests/tensorflow2/test_error_handling.py
```diff
@@ -0,0 +1,211 @@
...
+    logger = get_logger()
+    os.makedirs(out_dir)
+    file_handler = logging.FileHandler(filename=f"{out_dir}/tmp.log")
+    logger.addHandler(file_handler)
+    yield
```
connorgoggins, line 104: Can you explain the purpose of `yield` here?

connorgoggins (commented):
Looks good, thanks for making these additions! Just a few comments/questions.

smdebug/tensorflow/keras.py
ndodda-amazon: The scope of this change is to cover all smdebug functions called during training with the default smdebug configuration (from the AWS TF codebase essentially).

`wrap_optimizer` would only be called by the user manually and would thus be considered a custom smdebug configuration, so we don't wrap `wrap_optimizer` with the error handler.

tests/tensorflow2/test_error_handling.py
ndodda-amazon: The purpose of this fixture is to simulate setup and teardown functionality for each test in the file. hence the `autouse=True` in `pytest.fixture` - we don't need to manually reference this fixture in each test. 

There are three parts to this design:
- The block of code before `yield` is the setup and gets run before the test is executed.
- `yield` returns execution back to the test. The test then gets executed.
- The block of code after `yield` is the teardown and gets run after the test is executed (regardless of pass or fail).

This is a really interesting [read](https://docs.pytest.org/en/stable/fixture.html#teardown-cleanup-aka-fixture-finalization) that I'd recommend for more detail.

smdebug/core/hook.py
ndodda-amazon: Done.

smdebug/tensorflow/keras.py
ndodda-amazon: Clarified this in the PR description too.

connorgoggins: I see, thanks for clarifying.

connorgoggins (approved):

smdebug/core/error_handler.py
```diff
@@ -0,0 +1,132 @@
...
+            """
+            Set the hook to be used by the error handler. The hook is used to determine whether the ongoing training is
+            using the default smdebug configuration or a custom smdebug configuration.
+
+            This is meant to be called in the constructor of the relevant hook used for training. If an error occurs before
+            this function is called, the error handler will catch the error.
+            """
+            self.hook = hook
+
+        def catch_smdebug_errors(self, return_type=None, **handler_kwargs):
```
NihalHarish, line 54: Add return_val here

smdebug/core/error_handler.py
```diff
@@ -0,0 +1,132 @@
...
+            """
+
+            def wrapper(func):
+                @functools.wraps(func)
+                def error_handler(*args, **kwargs):
+                    # Determine default return value based on the return type.
+                    if return_type == bool:
+                        return_val = False
+                    elif return_type == "layer_call":
+                        return_val = handler_kwargs["old_call_fn"](*args, **kwargs)
```
NihalHarish, line 93: Change to generic: `fn(*args, **kwargs)`

smdebug/core/error_handler.py
```diff
@@ -0,0 +1,132 @@
...
+                    elif return_type == "wrap_tape":
+                        return_val = args[1]
```
NihalHarish, line 95: Check if this can be removed with return_val

smdebug/core/error_handler.py
```diff
@@ -0,0 +1,132 @@
...
+                    else:
+                        return_val = None
+
+                    # Return immediately if smdebug is disabled.
+                    if self.disable_smdebug:
+                        if return_type == "tape":
```
NihalHarish, line 101: Check if this can be done in a cleaner way by either checking for fn names or using a fn attribute

smdebug/core/hook.py
```diff
@@ -359,7 +362,7 @@ def _get_collections_to_save_for_step(self) -> Set["Collection"]:
                 )
         return self._collections_to_save_for_step
 
-    def is_tensor_saved_for_step(self, tensor_name):
+    def _is_tensor_saved_for_step(self, tensor_name):
```
NihalHarish, line 365: irrelevant change

NihalHarish (changes requested):
Implementation looks good. Refactor code. I am yet to review tests.

smdebug/core/error_handler.py
ndodda-amazon: Done.

smdebug/core/error_handler.py
ndodda-amazon: Done.

smdebug/core/error_handler.py
ndodda-amazon: Rewrote `wrap_tape` so that `tape` could be passed directly as a `return_val`. Done.

smdebug/core/error_handler.py
ndodda-amazon: Not necessary anymore, modified so that all return_val arguments that are functions get called here.

smdebug/core/error_handler.py
```diff
@@ -0,0 +1,136 @@
...
+                    except Exception as e:
+                        # If an smdebug error occurred with the default configuration or it occurred before the
+                        # configuration can even be determined (i.e. the constructor), catch the error and log it.
+
+                        if self.hook is None or self.hook.has_default_configuration():
+                            self.logger.error(BASE_ERROR_MESSAGE)
+                            self.logger.exception(e)  # Log stack trace.
+                            self.disable_smdebug = True  # Disable smdebug
+
+                            if callable(default_return_val):
```
NihalHarish, line 121: Nice! I learned something new. Can you explain what happens here as a comment?

ndodda-amazon: Explained it in the docstring for the class, but will add a comment here too.

NihalHarish (changes requested):

tests/zero_code_change/test_tensorflow2_gradtape_integration.py
```diff
@@ -27,6 +27,68 @@ def get_keras_data():
...
+                labels = tf.one_hot(labels, depth=10)
+                with tf.GradientTape(persistent=True) as tape:
+                    logits = model(data, training=True)  # (32,10)
+                    loss_value = cce(labels, logits)
+                grads = tape.gradient(loss_value, model.variables)
+                opt.apply_gradients(zip(grads, model.variables))
+                acc = train_acc_metric(dataset_labels, logits)
+            log = "Epoch %d " % (epoch + 1)
+            log += "Accuracy %.4f" % train_acc_metric.result()
+            print(log)
```
connorgoggins, line 88: Maybe use `logging` instead of `print` statements for output in the test?

connorgoggins (commented):
Just one small suggestion- other than that, looks great!

tests/zero_code_change/test_tensorflow2_gradtape_integration.py
ndodda-amazon: Copied this from the original helper function. But sure, I can replace this with logging.

smdebug/core/hook.py
```diff
@@ -153,6 +154,7 @@ def __init__(
         profiler_config_parser: ProfilerConfigParser object
             if passed, use this profiler configuration. by default, set up a new profiler configuration here.
         """
+        error_handler.set_hook(self)
```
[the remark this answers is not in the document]
ndodda-amazon: Doesn't need to be executed first. The error handler is designed to handle errors even if the hook is not set yet (default behavior is to catch the error).

What exactly are you proposing to assert? That the hook is set for the error handler?

`assert error_handler.hook is self`

smdebug/tensorflow/keras.py
```diff
@@ -1295,37 +1321,51 @@ def save_tape_logs(self, model_inputs=None, outputs=None):
             self._log_unsupported_tensors_in_non_eager_execution()
 
     def wrap_tape(self, tape):
-        """
-        Wrapping your GradientTape with this method enables finding gradient tensors and optimizer
-        variables.
+        # Don't modify the original tape object. If an error occurs when wrapping the tape functions, we simply return
+        # the original tape object.
+        wrapped_tape = deepcopy(tape)
```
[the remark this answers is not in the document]
ndodda-amazon: Size isn't an issue with the tape object, since nested GradientTape is a valid use case. Doing a deep copy is the same as creating another GradientTape object with the same attributes.

Basically, we're copying the original tape object so that it purposefully stays unmodified. If an error occurs in wrapping the copy, we simply return the original tape object. The scope is only wrapping the tape, not the actual tape functions themselves.

tests/core/test_error_handling.py
```diff
@@ -0,0 +1,191 @@
...
+
+@pytest.fixture
+def value_error_message():
+    return (
+        "If this ValueError causes the test to fail, the error handler failed to catch the error!"
+    )
+
+
+@pytest.fixture
+def dummy_clean_function():
```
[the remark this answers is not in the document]
ndodda-amazon: Done

tests/zero_code_change/test_tensorflow2_gradtape_integration.py
ndodda-amazon: Done

connorgoggins (approved):

tests/core/test_error_handling_agent.py
```diff
@@ -0,0 +1,189 @@
...
+import pytest
+
+# First Party
+from smdebug.core.error_handling_agent import BASE_ERROR_MESSAGE
+from smdebug.core.logger import DuplicateLogFilter, get_logger
+from smdebug.core.utils import error_handling_agent
+
+
+@pytest.fixture
+def stack_trace_filepath(out_dir):
```
NihalHarish, line 16: Try to avoid the duplication of these fixtures.

tests/zero_code_change/test_tensorflow2_error_handling_agent.py
```diff
@@ -0,0 +1,406 @@
...
+    Set up each test to:
+        - Add a logging handler to write all logs to a file (which will be used to verify caught errors in the tests)
+        - Remove the duplicate logging filter
+        - Reset the error handling agent after the test so that smdebug is reenabled.
+    """
+    old_create_from_json = Hook.create_from_json_file
+    del_hook()
+
+    logger = get_logger()
+    os.makedirs(out_dir)
```
NihalHarish, line 274: You can move these repetitive operations to the fixture

tests/zero_code_change/test_tensorflow2_error_handling_agent.py
```diff
@@ -0,0 +1,406 @@
...
+    file_handler = logging.FileHandler(filename=stack_trace_filepath)
+    logger.addHandler(file_handler)
+    duplicate_log_filter = None
+    for log_filter in logger.filters:
+        if isinstance(log_filter, DuplicateLogFilter):
+            duplicate_log_filter = log_filter
+            break
+    logger.removeFilter(duplicate_log_filter)
+
+    yield
```
NihalHarish, line 284: What purpose do these `yield` statements serve if they are outside of the test function.

NihalHarish (changes requested):

tests/zero_code_change/test_tensorflow2_error_handling_agent.py
ndodda-amazon: These operations are in the fixture?

tests/zero_code_change/test_tensorflow2_error_handling_agent.py
ndodda-amazon: https://github.com/awslabs/sagemaker-debugger/pull/483#discussion_r620699789

smdebug/tensorflow/keras.py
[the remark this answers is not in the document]
ndodda-amazon: Done

tests/core/test_error_handling_agent.py
ndodda-amazon: Done

tests/core/test_error_handling_agent.py
```diff
@@ -12,11 +12,6 @@
 from smdebug.core.utils import error_handling_agent
 
 
-@pytest.fixture
-def stack_trace_filepath(out_dir):
-    return f"{out_dir}/tmp.log"
-
-
 @pytest.fixture
 def runtime_error_message():
```
NihalHarish, line 16: Parameterize this string

NihalHarish (changes requested):

tests/core/test_error_handling_agent.py
ndodda-amazon: Done.

NihalHarish (approved):

PR #484, opened 2021-04-29
Implement error handling for PyTorch

ndodda-amazon:
### Description of changes:
Followup to #483 to implement error handler for PyTorch. Since #483 is not merged yet, this PR technically contains changes for the error handler + BaseHook too.

EDIT: #483 has been merged, so this diff only has changes unique to error handling for PyTorch.

#### Style and formatting:


tests/core/test_error_handling.py
```diff
@@ -0,0 +1,191 @@
...
+    def raise_error():
+        raise ValueError(value_error_message)
+
+    return raise_error
+
+
+@pytest.fixture(autouse=True)
+def set_up_logging_and_error_handler(out_dir, stack_trace_filepath):
+    """
+    Setup up each test to:
```
connorgoggins, line 73: Drop "up", should be "Set up each test to".

tests/core/test_error_handling.py
```diff
@@ -0,0 +1,191 @@
...
+        - Add a logging handler to write all logs to a file (which will be used to verify caught errors in the tests)
+        - Remove the duplicate logging filter
+        - Reset the error handler after the test so it that it is reenabled.
```
connorgoggins, line 76: Should be "Reset the error handler after the test so it is reenabled".

tests/zero_code_change/pt_utils.py
```diff
@@ -5,10 +5,14 @@
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
+import torch.optim as optim
 import torchvision
 import torchvision.transforms as transforms
 from packaging import version
 
+# First Party
+import smdebug.pytorch as smd
```
connorgoggins, line 14: nit (non-blocking): maybe use a different variable name to make it clear that this is the PyTorch module of smdebug?

tests/zero_code_change/pt_utils.py
```diff
@@ -58,3 +62,36 @@ def forward(self, x):
...
+            loss = criterion(outputs, labels)
+        else:
+            loss = F.cross_entropy(outputs, labels)
+            if script_mode:
+                hook.record_tensor_value(tensor_name="loss", tensor_value=loss)
+        loss.backward()
+        optimizer.step()
+
+        if i == 499:  # print every 2000 mini-batches
+            break
```
connorgoggins, line 97: I'm confused by this `if` block and the corresponding comment - how does breaking on the 500th iteration of the loop correspond to printing every 2000 mini-batches? My guess is that this comment is outdated.

tests/zero_code_change/test_pytorch_error_handling.py
```diff
@@ -0,0 +1,353 @@
...
+            monkeypatch.setenv("SMPROFILER_CONFIG_PATH", profiler_config_path)
+            return HookWithBadTorchCallbackAndCustomProfilerConfiguration(out_dir=out_dir)
+
+    return HookWithBadTorchCallbackAndCustomProfilerConfiguration
+
+
+@pytest.fixture(autouse=True)
+def set_up_logging_and_error_handler(out_dir, stack_trace_filepath):
+    """
+    Setup up each test to:
```
connorgoggins, line 215: Change docstring in the same way as described in the first comment.

connorgoggins (commented):
Great work, looks good overall! Just a few small comments.

tests/zero_code_change/pt_utils.py
ndodda-amazon: Agreed but this is consistent with the way all of the imports of the framework module are done. Unless we take a decision to refactor all of the files in smdebug and the AWS framework codebases, it's better to just be consistent with the naming.

tests/zero_code_change/pt_utils.py
ndodda-amazon: Not sure what the intention of the comment is (I just refactored the code). Will remove the comment since I agree it's outdated.

ndodda-amazon: Done.

tests/core/test_error_handling.py
ndodda-amazon: Done.

tests/core/test_error_handling.py
ndodda-amazon: Done.

tests/zero_code_change/test_pytorch_error_handling.py
ndodda-amazon: Done.

leleamol (changes requested):
The changes are looking good. Can you post the performance impact with and without these changes?

leleamol (approved):

smdebug/pytorch/hook.py
```diff
@@ -586,6 +593,7 @@ def count_parameters(self, model):
         self.logger.info(f"Total Trainable Params: {total_params}")
         return total_params
 
+    @error_handling_agent.catch_smdebug_errors()
```
NihalHarish, line 596: What is the behavior of this decorator for recursive calls?
How many logs will it print?

ndodda-amazon: Interesting question.

The error handler is designed so that:
- If smdebug is already disabled, no new functions will be called (and a default return value is returned)
- When a new error is caught, the error is logged. This logging behavior is the same as any other logging, in that there is a duplicate log filter present.

So let's say the error handler is wrapped around a function with recursive calls. Let's say the first iteration is n1 and the fifth iteration is n5. If an error occurs at n5, a default return value is returned at that point, and potentially more errors could be caught at n4 and above. There isn't really an easy solution for this issue, we should try not to have any wrapped functions that have recursive calls.

However, the log filter will always be in effect. So there are two scenarios:
- Errors at n4 and above are the same as n5. Log filter limits the number of logged errors to 3.
- Errors at n4 and above are different from n5. Then these errors messages are different and should be logged (indicates multiple smdebug errors).

PR #492, opened 2021-05-07
Skip Telemetry If Running Inside Papermill

NihalHarish:
### Description of changes:
- The `setup_profiler_report` now only executes outside of papermill by checking for an env variable
#### Nits
- Docstrings for telemetry functions

#### Style and formatting:


smdebug/core/utils.py
```diff
@@ -102,6 +111,13 @@ def prepare_telemetry_url(processing_job_arn):
 
 
 def setup_profiler_report(processing_job_arn, opt_out=False):
+    """
```
ndodda-amazon, line 114: What is the TODO?

smdebug/core/utils.py
```diff
@@ -102,6 +111,13 @@ def prepare_telemetry_url(processing_job_arn):
...
+    TODO
```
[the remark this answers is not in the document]
NihalHarish: This was a wip pr

smdebug/core/utils.py
```diff
@@ -102,11 +112,24 @@ def prepare_telemetry_url(processing_job_arn):
...
+    in which case, we skip this skip the telemetry request.
+    :param processing_job_arn:
+    :param opt_out: Does not make a telemetry call if opt_out=True
+    Makes a GET request to the prepared telemetry url if opt_out=False
+    :return:
+    """
+    if (
+        opt_out is False
+        and bool(processing_job_arn)
+        and not bool(os.getenv(PAPERMILL_EXECUTION_ENV_VAR))
```
ndodda-amazon, line 128: Where will this env variable normally be set? I'd suggest we do it here so that it's all managed within this function.

tests/core/test_utils.py
```diff
@@ -271,14 +277,47 @@ def _get_all_aws_regions():
...
+    return arn
+
+
+def false_get(*args, **kwargs):
+    assert False, "GET Should Not Be Called"
+
+
+def true_get(*args, **kwargs):
+    print("GET Should Be Called")
+    assert True
```
ndodda-amazon, line 292: nit: Can just make this function definition empty.

tests/core/test_utils.py
```diff
@@ -271,14 +277,47 @@ def _get_all_aws_regions():
...
+
+def test_setup_profiler_report(monkeypatch, test_arn):
+    test_arn = test_arn.format(region="us-east-1")
+    monkeypatch.setattr(requests, "get", true_get)
+
+    # setup_profiler_report is expected to be executed when called with a correct ARN
+    setup_profiler_report(test_arn)
+
+    # setup_profiler_report is expected to be executed when called with opt_out=False
+    setup_profiler_report(test_arn, opt_out=False)
```
ndodda-amazon, line 315: How are you verifying that `setup_profiler_report` was called? Not sure if this is possible with monkeypatch, but using the `@patch` decorator, you can check the number of calls made to a mocked function. So one idea is to "mock" the `requests.get` function as you've already done so you can simply check the number of times it was called.

[Example](https://stackoverflow.com/a/21983089)

ndodda-amazon (changes requested):
Changes look good. How are you verifying that telemetry is skipped when running inside papermill for these changes?

smdebug/core/utils.py
NihalHarish: This env is set in the by the rule executing papermill.
Why would this fn set it?

tests/core/test_utils.py
NihalHarish: monkeypatch does exactly that. If `false_get` is executed, the test fails

smdebug/core/utils.py
```diff
@@ -102,11 +112,24 @@ def prepare_telemetry_url(processing_job_arn):
 
 
 def setup_profiler_report(processing_job_arn, opt_out=False):
-    # This function is used externally in the profiler report
-    if opt_out is False and bool(processing_job_arn):
-        prepared_telemtry_url = prepare_telemetry_url(processing_job_arn)
+    """
+    This function is used externally in the profiler report
+    We check for the if the env variable "NOTEBOOK_EXECUTION" has been set to determine
```
NihalHarish, line 117: nit: update comment

tests/core/test_utils.py
```diff
@@ -271,14 +277,51 @@ def _get_all_aws_regions():
...
+
+
+def test_setup_profiler_report(monkeypatch, test_arn):
+    test_arn = test_arn.format(region="us-east-1")
+    monkeypatch.setattr(requests, "get", fake_get)
+
+    # setup_profiler_report is expected to be executed when called with a correct ARN
+    try:
+        setup_profiler_report(test_arn)
+    except AssertionError:
```
ndodda-amazon, line 310: same problem - if the function was not called, no error is caught and execution continues. but you're on the right track - you can actually assert that an error was thrown.

Docs: https://docs.pytest.org/en/doc-warnings/assert.html#assertions-about-expected-exceptions
Example from error handling agent tests: https://github.com/awslabs/sagemaker-debugger/blob/master/tests/zero_code_change/test_tensorflow2_error_handling_agent.py#L390

tests/core/test_utils.py
```diff
@@ -271,14 +277,51 @@ def _get_all_aws_regions():
...
+        # get is expected to be called
+        pass
+
+    # setup_profiler_report is expected to be executed when called with opt_out=False
+    try:
+        setup_profiler_report(test_arn, opt_out=False)
```
ndodda-amazon, line 316: same as above

tests/core/test_utils.py
```diff
@@ -271,14 +277,51 @@ def _get_all_aws_regions():
...
+    except AssertionError:
+        # get is expected to be called
+        pass
+
+    monkeypatch.setattr(requests, "get", fake_get)
```
ndodda-amazon, line 321: Don't need this line anymore. If you assert that an error was thrown for the above calls, then normally calling the below two functions with `fake_get` implicitly tests that the get call was not made, otherwise there would have been an assertion error.

ndodda-amazon (changes requested):

tests/core/test_utils.py
NihalHarish: Done

tests/core/test_utils.py
NihalHarish: Done

tests/core/test_utils.py
NihalHarish: Thanks! Done

ndodda-amazon (approved):
LGTM, great work! Feel free merge after the CI is green.

PR #497, opened 2021-05-14
Add safety checks for TF2 error handling for GradientTape

ndodda-amazon:
### Description of changes:
Adding safety checks to the TF2 error handling for GradientTape:
- If smdebug is disabled and `push_tape` is called, only call the original `push_tape` function if the tape isn't already recording.
- If smdebug is disabled and `pop_tape` is called, only call the original `pop_tape` function if the tape is already recording.

These changes serve as stronger verification to ensure that the error handling agent does not throw an error while calling original tape functions when smdebug is disabled.

#### Style and formatting:

### DIFF
--- a/smdebug/core/error_handling_agent.py
+++ b/smdebug/core/error_handling_agent.py
@@ -49,7 +49,7 @@ class ErrorHandlingAgent(object):
             """
             self.hook = hook
 
-        def catch_smdebug_errors(self, default_return_val=None, return_type=None):
+        def catch_smdebug_errors(self, default_return_val=None):
             """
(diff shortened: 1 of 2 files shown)


smdebug/core/error_handling_agent.py
```diff
@@ -49,7 +49,7 @@ def set_hook(self, hook):
             """
             self.hook = hook
 
-        def catch_smdebug_errors(self, default_return_val=None, return_type=None):
+        def catch_smdebug_errors(self, default_return_val=None):
```
ndodda-amazon, line 52: Irrelevant change - removing a parameter that doesn't get used.

NihalHarish (approved):

PR #499, opened 2021-05-20
Add error handling for MXNet

ndodda-amazon:
### Description of changes:
Add error handling for MXNet. Wrapped the `register_block`, `forward_hook` and `forward_pre_hook` functions, which get called directly from MXNet.

#### Style and formatting:

### DIFF
--- a/tests/zero_code_change/test_pytorch_error_handling_agent.py
+++ b/tests/zero_code_change/test_pytorch_error_handling_agent.py
@@ -305,7 +305,7 @@ def test_pytorch_error_handling(
         assert stack_trace_logs.count(error_message) == 2
 
         # `fhook` should not have errored if `register_module` already errored.
-        if hook_type == "keras_and_layer_callback_error":
+        if hook_type == "torch_callback_and_register_module_error":
             assert torch_callback_error_message not in stack_trace_logs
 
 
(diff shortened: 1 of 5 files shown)


tests/zero_code_change/test_pytorch_error_handling_agent.py
```diff
@@ -305,7 +305,7 @@ def test_pytorch_error_handling(
         assert stack_trace_logs.count(error_message) == 2
 
         # `fhook` should not have errored if `register_module` already errored.
-        if hook_type == "keras_and_layer_callback_error":
+        if hook_type == "torch_callback_and_register_module_error":
```
ndodda-amazon, line 308: Irrelevant change, fixed a bug in a test.

NihalHarish: How did this pass CI the first time?

NihalHarish (approved):

tests/zero_code_change/test_pytorch_error_handling_agent.py
ndodda-amazon: This checks a condition and then makes an assertion. So it "passed" in the CI the first time because the condition was always False for the corresponding test, hence the additional assertion was never made.

PR #500, opened 2021-05-25
Add support for Python profiling with native TF2 training + refactoring Python profiling

ndodda-amazon:
### Description of changes:
With this change, users have the option of enabling profiling in their training script with a new context manager `hook.profiler`. See docstring and tests for information on usage.

Currently, this change only addresses Python profiling support for native TF2.

TODO: Add profiling support for the other types of profiling for native TF2.

#### Style and formatting:


smdebug/tensorflow/keras.py
```diff
@@ -122,9 +124,12 @@ def __init__(
         self.has_logged_unsupported_tensors_in_non_eager_execution = False
         self.prepared_tf2_collections = False
         self.prepared_gradient_tape_collections = False
+        # this flag indicates whether profiling is enabled for native TF2 training
+        self.is_profiler_enabled_for_native_training = False
+        self.closed = False
```
NihalHarish, line 129: Rename `self.closed` to something more descriptive.

smdebug/tensorflow/keras.py
```diff
@@ -736,6 +741,46 @@ def _prepare_collections_for_tf2(self):
             self._unwrap_model_with_input_output_saver()
         self.prepared_tf2_collections = True
 
+    def _handle_start_step_python_profiling(self, mode, current_step):
+        """
+        This function is called to handle python profiling at the start of a step.
+        :param mode: ModeKeys.TRAIN ModeKeys.EVAL ModeKeys.PREDICT
+        """
+        if self.python_profiler:
+            self.python_profiler.stop_profiling(
```
NihalHarish, line 750: Handle the if-case inside of `stop_profiling`

smdebug/tensorflow/keras.py
```diff
@@ -736,6 +741,46 @@ def _prepare_collections_for_tf2(self):
...
+                StepPhase.STEP_START,
+                end_mode=mode_keys_to_python_profile_mode(mode),
+                end_step=current_step,
+            )
+            if self.profiler_config_parser.should_save_metrics(
+                MetricsCategory.PYTHON_PROFILING, current_step
+            ):
+                self.python_profiler.start_profiling(
```
NihalHarish, line 758: Re-evaluate the profiler can accept the profiling config and handle all the cases.

NihalHarish (changes requested):
Changes look good. Can you refactor the code based on my suggestions?

smdebug/tensorflow/keras.py
```diff
@@ -122,9 +124,12 @@ def __init__(
...
         self.prepared_tf2_collections = False
         self.prepared_gradient_tape_collections = False
+        # this flag indicates whether profiling is enabled for native TF2 training
+        self.is_profiler_enabled_for_native_training = False
+        self.closed = False
 
-        if python_profiler:
-            atexit.register(python_profiler.stop_profiling, StepPhase.END)
+        if self.python_profiler:
+            atexit.register(self.profiling_end)
```
leleamol, line 132: At the end of program, will this object (self) be still valid?  What happens if it goes out of scope?

ndodda-amazon: It's still valid. `self.profiling_end` is passed as an argument like any other `callable` in Python, so it can directly be called without `self` needing to be in scope. 

As verification that this works, the end to end SageMaker tests pass (with the post-hook-close Python profiling stats being successfully collected for the jobs).

smdebug/tensorflow/keras.py
ndodda-amazon: Done

smdebug/tensorflow/keras.py
ndodda-amazon: Done

smdebug/tensorflow/keras.py
ndodda-amazon: Done

smdebug/profiler/profiler_config_parser.py
```diff
@@ -332,19 +339,83 @@ def write_tf_dataloader_flag(self, flag_filename):
...
+                step_phase,
+                start_mode=mode_keys_to_python_profile_mode(mode),
+                start_step=current_step,
+            )
+
+    def start_pre_step_zero_python_profiling(self):
+        """Start pre-step zero python profiling if python profiling is enabled.
+        """
+        if not self.profiling_enabled or not self.config.python_profiling_config.is_enabled():
+            return
```
NihalHarish, line 365: nit: encapsulate this boolean with a function to reduce duplication.

smdebug/profiler/profiler_config_parser.py
```diff
@@ -332,19 +339,83 @@ def write_tf_dataloader_flag(self, flag_filename):
...
+    ProfilerConfig object if it doesn't already exist. Takes in the framework being used as an argument.
+
+    If `create_new` is set to `True`, then a new profiler config parser will be created regardless of whether one
+    already exists or not.
     :return:
     """
     global _profiler_config_parser
-    if _profiler_config_parser is None:
-        _profiler_config_parser = ProfilerConfigParser()
+    if create_new or _profiler_config_parser is None:
```
NihalHarish, line 417: nit: Rename `create_new` to `should_create_new_profiler_config_parser`

NihalHarish: What does `ProfilerConfigParser(framework=framework, create_new=create_new)` return for the following conditions?

```
| create_new | profiler_config |
|------------|-----------------|
| True       | True            |
| True       | False           |
| False      | True            |
| False      | False           |
```


I think there is scope to reduce this code to just

```
return ProfilerConfigParser(framework=framework, create_new=create_new)
```

smdebug/profiler/profiler_config_parser.py
```diff
@@ -356,3 +427,5 @@ def reset_profiler_config_parser():
     """
     global _profiler_config_parser
     _profiler_config_parser = None
+
+    reset_python_profiler()
```
NihalHarish, line 431: if this function is only intended to be called by the test framework, can we indicate that with the fn name?
Maybe with `_` as a prefix?

NihalHarish (changes requested):

smdebug/profiler/profiler_config_parser.py
ndodda-amazon: We should only prefix with `_` for functions called in the same module. 

But perhaps we can move this reset function to the test utils (along with the Python profiling one)

ndodda-amazon: We should only prefix with `_` if the function is only used inside the same module. This function gets called outside the module.

smdebug/profiler/profiler_config_parser.py
ndodda-amazon: > I think there is scope to reduce this code to just
> 
> ```
> return ProfilerConfigParser(framework=framework, create_new=create_new)
> ```

no, the point is that if `create_new` is True and `profiler_config_parser` isn't initialized, we should create a new profiler config parser. otherwise we return the existing profiler config parser.

NihalHarish: If create_new is True we return a new ProfilerConfigParser regardless of the value of profiler_config_parser, right?

smdebug/profiler/profiler_config_parser.py
ndodda-amazon: Done.

smdebug/profiler/profiler_config_parser.py
ndodda-amazon: Added to `reset_python_profiler`'s docstring to make it clear that it is only intended to be used for testing.

smdebug/profiler/profiler_config_parser.py
ndodda-amazon: But if `create_new` is False, then we shouldn't return a new ProfilerConfigParser object if an object already exists.

smdebug/tensorflow/context_manager.py
```diff
@@ -0,0 +1,10 @@
+class ProfilerContextManager(object):
```
NihalHarish, line 1: nit: rename file to `profiler_context_manager`

smdebug/tensorflow/context_manager.py
```diff
@@ -0,0 +1,10 @@
...
+    def __init__(self, hook, mode):
+        self.hook = hook
+        self.mode = mode
+
+    def __enter__(self):
+        self.hook.profiling_start_batch(self.mode)
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.hook.profiling_end_batch(self.mode)
```
NihalHarish, line 10: Check if we can eliminate the need for a separate class for a context manager?

smdebug/tensorflow/keras.py
```diff
@@ -1383,3 +1332,80 @@ def record_tensor_value(self, tensor_name, tensor_value):
...
+                    opt.apply_gradients(zip(grads, model.variables))
+                cleanup()
+        ```
+        """
+        _ = self.wrap_tape(
+            tf.GradientTape()
+        )  # GradientTape functions must be wrapped to correctly manage current step
+        self.set_mode(mode)
+        self.is_profiler_enabled_for_native_training = True
+        return ProfilerContextManager(self, mode)
```
NihalHarish, line 1411: See if you can use a `yield` statement instead of managing the context manager with an object.

NihalHarish (changes requested):

tests/profiler/tensorflow2/test_native_tf2_profiler.py
```diff
@@ -0,0 +1,193 @@
...
+    python_stats_dir = os.path.join(out_dir, "framework", "tensorflow", profiler_name)
+
+    assert os.path.isdir(python_stats_dir)
+
+    for node_id in os.listdir(python_stats_dir):
+        node_dir_path = os.path.join(python_stats_dir, node_id)
+        stats_dirs = os.listdir(node_dir_path)
+
+        for stats_dir in stats_dirs:
+            print(stats_dir)
```
NihalHarish, line 135: Remove this

tests/profiler/tensorflow2/test_native_tf2_profiler.py
```diff
@@ -0,0 +1,193 @@
...
+
+        # The expected number of stats directories during is (num_steps * 2) + 1. This includes profiling for both
+        # phases of each step and pre-step zero python profiling and post-hook-close python profiling.
+        assert (
+            len(stats_dirs)
+            == profiler_config_parser.config.python_profiling_config.num_steps * 2 + 2
```
NihalHarish, line 141: nit: `pemdas`

tests/profiler/tensorflow2/test_native_tf2_profiler.py
```diff
@@ -0,0 +1,193 @@
...
+                stats_path = os.path.join(stats_dir_path, stats_file)
+                if stats_file == CPROFILE_STATS_FILENAME:
+                    assert pstats.Stats(stats_path)
+                elif stats_file == PYINSTRUMENT_JSON_FILENAME:
+                    with open(stats_path, "r") as f:
+                        assert json.load(f)
+
+
+@pytest.mark.parametrize("python_profiler_name", [CPROFILE_NAME, PYINSTRUMENT_NAME])
+@pytest.mark.parametrize("model_type", ["sequential", "functional", "subclassed"])
```
NihalHarish, line 161: enum

NihalHarish (changes requested):

smdebug/profiler/profiler_config_parser.py
```diff
@@ -332,19 +341,83 @@ def write_tf_dataloader_flag(self, flag_filename):
...
+        """Handle python profiling at the end of the step with the given mode and step by stopping python profiling
+        and starting python profiling again if python profiling stats should be saved for the current step.
+        """
+        return self._handle_step_python_profiling(StepPhase.STEP_END, mode, current_step)
+
+    def handle_forward_pass_end_python_profiling(self, mode: ModeKeys, current_step):
+        """Handle python profiling at the end of the forward pass with the given mode and step by stopping python
+        profiling and starting python profiling again if python profiling stats should be saved for the current step.
+        """
+        return self._handle_step_python_profiling(StepPhase.FORWARD_PASS_END, mode, current_step)
```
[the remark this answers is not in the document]
ndodda-amazon: Hooks shouldn't have any internal knowledge of python profiling and how it defines step phase, since it's framework agnostic. This should only fall under the scope of the profiler config parser, which will be the only component to have direct interaction with python profiling.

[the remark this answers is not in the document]
ndodda-amazon: At least, that's the purpose of the design change in this PR.

smdebug/tensorflow/context_manager.py
ndodda-amazon: Will remove the class and just use `contextlib.context_manager`

smdebug/profiler/profiler_config_parser.py
```diff
@@ -332,19 +341,83 @@ def write_tf_dataloader_flag(self, flag_filename):
...
+        if not self.is_python_profiling_enabled():
+            return
+
+        return self.python_profiler.start_profiling(StepPhase.END)
+
+    def stop_post_hook_close_python_profiling(self):
+        """Stop post-hook-close python profiling if python profiling is enabled.
+        """
+        if not self.profiling_enabled or not self.config.python_profiling_config.is_enabled():
+            return
```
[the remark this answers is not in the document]
ndodda-amazon: Done

smdebug/tensorflow/keras.py
```diff
@@ -1383,3 +1332,80 @@ def record_tensor_value(self, tensor_name, tensor_value):
...
+
+        self.profiler_config_parser.load_config()
+        self.profiler_config_parser.handle_step_start_python_profiling(mode, current_step)
+
+    def profiling_end_batch(self, mode=ModeKeys.TRAIN):
+        """
+        Enabling profiler at the end of train batch for native Tf2 training.
+        :param mode: ModeKeys.TRAIN ModeKeys.EVAL ModeKeys.PREDICT
+
+        TODO: Add support for other profiling at the end of the batch.
```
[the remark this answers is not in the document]
ndodda-amazon: Done

smdebug/tensorflow/keras.py
ndodda-amazon: Done

tests/profiler/tensorflow2/test_native_tf2_profiler.py
ndodda-amazon: Done

tests/profiler/tensorflow2/test_native_tf2_profiler.py
ndodda-amazon: Added parentheses

tests/profiler/tensorflow2/test_native_tf2_profiler.py
ndodda-amazon: Done

PR #501, opened 2021-06-09
Enabling the support for Pytorch 1.9

leleamol:
### Description of changes:
The change enables support for PyTorch 1.9

The autograd based detailed profiling can not be supported .
Ran the integration tests with PT1.9 binary.

#### Style and formatting:

### DIFF
--- a/smdebug/pytorch/utils.py
+++ b/smdebug/pytorch/utils.py
@@ -82,11 +82,11 @@ def is_pt_1_8():
 @lru_cache(maxsize=1)
 def is_pt_1_9():
     """
-    Determine whether the version of torch is 1.8.x
+    Determine whether the version of torch is 1.9.x
     :return: bool
     """
-    return version.parse("1.9.0") <= PT_VERSION < version.parse("2.0.0")
+    return version.parse("1.9.0") <= PT_VERSION < version.parse("1.10.0")
 
 
 def is_current_version_supported(pytorch_version=torch.__version__):
-    return version.parse("1.5.0") <= version.parse(pytorch_version) < version.parse("2.0.0")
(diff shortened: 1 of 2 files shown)


smdebug/pytorch/utils.py
```diff
@@ -79,5 +79,14 @@ def is_pt_1_8():
     return version.parse("1.8.0") <= PT_VERSION < version.parse("1.9.0")
 
 
+@lru_cache(maxsize=1)
+def is_pt_1_9():
+    """
+    Determine whether the version of torch is 1.8.x
+    :return: bool
+    """
+    return version.parse("1.9.0") <= PT_VERSION < version.parse("2.0.0")
```
ndodda-amazon, line 88: nit: version numbers can technically go over 9, so we should check instead that:
```
return version.parse("1.5.0") <= PT_VERSION < version.parse("1.10.0")
```

smdebug/pytorch/utils.py
```diff
@@ -79,5 +79,14 @@ def is_pt_1_8():
...
+
+
 def is_current_version_supported(pytorch_version=torch.__version__):
-    return version.parse("1.5.0") <= version.parse(pytorch_version) < version.parse("1.9.0")
+    return version.parse("1.5.0") <= version.parse(pytorch_version) < version.parse("2.0.0")
```
ndodda-amazon, line 92: Same as above.

ndodda-amazon (changes requested):
Looks good, left minor comments.

smdebug/pytorch/utils.py
leleamol: It can go either way. In any case we would have to update this line when a new version comes up

ndodda-amazon (approved):
LGTM, feel free to merge after the CI is green.

PR #509, opened 2021-06-30
smprofiler: update test requirements, add pip check statement to check for broken requirements

arjkesh:
### Description of changes:
SM profiler requirements have conflicts that result in nonzero exit code when running "pip check" - updating the requirements for now.

Added boto3/botocore to ensure that awscli/boto3/botocore don't go out of sync in any downstream dependencies.

Long term fix for this will be to use a package dependency manager.

#### Style and formatting:


config/profiler/requirements.txt
```diff
@@ -1,12 +1,14 @@
...
-pyYaml==5.4
-pytest-html==3.0.0
-sagemaker==2.23.0
-pre-commit==2.6.0
-awscli==1.18.203
-pytest==6.1.2
-pytest-cov==2.10.1
+awscli==1.19.102
+boto3==1.17.102
+botocore==1.20.102
```
ndodda-amazon, line 4: The tests shouldn't manage dependencies for `boto3` and `botocore` since the tests don't consume them directly. `sagemaker` and `awscli` will manage them as dependencies, so let's remove these two packages.

config/profiler/run_profiler_integration_tests.sh
```diff
@@ -54,6 +54,9 @@ apt-get install sudo -qq -o=Dpkg::Use-Pty=0 >/dev/null 2>/dev/null # mask output
 sudo apt-get install unzip -qq -o=Dpkg::Use-Pty=0 >/dev/null 2>/dev/null # mask output
 pip install -q -r config/profiler/requirements.txt >/dev/null 2>/dev/null # mask output
 
+# Check for broken dependencies
+pip check
```
ndodda-amazon, line 58: This checks for broken dependencies but doesn't actually halt the build if there are broken dependencies.

Check the exit code of this call, print an appropriate message, and then call `exit 1`

ndodda-amazon (changes requested):
Looks good, left some minor comments.

ndodda-amazon (changes requested):
Can we temporarily comment out lines 48-50? It looks like the logic to determine whether the profiler integration tests in the CI breaks when the source branch doesn't belong to this repository.

After we confirm those tests pass, we can revert the changes and merge.

arjkesh:
> Can we temporarily comment out lines 48-50 in `run_profiler_integration_tests.sh`? It looks like the logic to determine whether the profiler integration tests in the CI breaks when the source branch doesn't belong to this repository.
> 
> After we confirm those tests pass, we can revert the changes and merge.

confirmed tests pass in c12549c00b4ece3efee0a12192818d280ffda608

ndodda-amazon (approved):
LGTM, thanks for making this change!

PR #513, opened 2021-07-10
Install Protoc With setup.py

NihalHarish:
### Description of changes:
- updated `setup.py` to install protoc

#### Style and formatting:

### DIFF
--- a/compile_protobuf.py
+++ b/compile_protobuf.py
@@ -48,7 +48,11 @@ def get_protoc_download_url():
         archive_url = f"https://github.com/protocolbuffers/protobuf/releases/download/v3.11.4/protoc-3.11.4-linux-{machine}.zip"
         logging.info("Downloading protoc for Linux: %s", archive_url)
     else:
-        archive_url = None
+        system = platform.system()
+        raise RuntimeError(
+            f"Could not find protoc for System: {system} Machine: {machine}.\
+         Please install it manually by running sh protoc_downloader.sh"
+        )
     return archive_url
 
 
@@ -59,11 +63,6 @@ def get_protoc():
         return shutil.which("protoc"), None
 
     archive_url = get_protoc_download_url()
-    if not archive_url:
-        system = platform.system()
-        raise RuntimeError(
-            f"protoc not installation on {system}. Please install it manually by running sh protoc_downloader.sh"
-        )
     (fname, headers) = urllib.request.urlretrieve(archive_url)
     tmpdir = tempfile.mkdtemp(prefix="protoc_smdebug")
     with ZipFile(fname, "r") as zipf:
--- a/compile_protobuf.py
+++ b/compile_protobuf.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# Standard Library
+import glob
+import logging
+import os
+import shutil
+import sys
+import tempfile
+import urllib
+import urllib.request
+from subprocess import check_call
+from zipfile import ZipFile
+
+
+def script_name() -> str:
+    """:returns: script name with leading paths removed"""
+    return os.path.split(sys.argv[0])[1]
+
+
+def config_logging():
+    import time
+
+    logging.getLogger().setLevel(logging.INFO)
+    logging.basicConfig(format="{}: %(asctime)sZ %(levelname)s %(message)s".format(script_name()))
+    logging.Formatter.converter = time.gmtime
+
+
+def _protoc_bundle():
+    """
+    Returns an archive with the binary protoc distro for the platform
+    """
+    import platform
+
+    system = platform.system()
+    machine = platform.machine()
+    if system == "Darwin":
+        archive_url = "https://github.com/protocolbuffers/protobuf/releases/download/v3.11.4/protoc-3.11.4-osx-x86_64.zip"
+        logging.info("Downloading protoc for Darwin: %s", archive_url)
+    elif system == "Linux":
+        archive_url = f"https://github.com/protocolbuffers/protobuf/releases/download/v3.11.4/protoc-3.11.4-linux-{machine}.zip"
+        logging.info("Downloading protoc for Linux: %s", archive_url)
+    else:
+        archive_url = None
+    return archive_url
+
+
+def get_protoc():
+    """make sure protoc is available, otherwise download it and return a tuple with the protoc
+    binary and a temporary dir if it needed to be downloaded"""
+    if not shutil.which("protoc"):
+        archive_url = _protoc_bundle()
+        if not archive_url:
+            raise RuntimeError(
+                "protoc not installed. Please install it manually by running sh protoc_downloader.sh"
+            )
+        logging.info("Downloading protoc")
+        (fname, headers) = urllib.request.urlretrieve(archive_url)
+        tmpdir = tempfile.mkdtemp(prefix="protoc_smdebug")
+        with ZipFile(fname, "r") as zipf:
+            zipf.extractall(tmpdir)
+        protoc_bin = os.path.join(tmpdir, "bin", "protoc")
+        os.chmod(protoc_bin, 0o755)
+        return (protoc_bin, tmpdir)
+    return (shutil.which("protoc"), None)
+
+
+def compile_protobuf():
+    """
+    Compile protobuf files for smdebug
+    """
+    logging.info("Compile protobuf")
+    logging.info("================")
+    (protoc_bin, tmpdir) = get_protoc()
+    cmd = [protoc_bin]
+    proto_files = glob.glob("smdebug/core/tfevent/proto/*.proto")
+    cmd.extend(proto_files)
+    cmd.append("--python_out=.")
+    logging.info("Call to protoc: %s", " ".join(cmd))
+    check_call(cmd)
+    if tmpdir:
+        shutil.rmtree(tmpdir, ignore_errors=True)
+
+
+def main():
+    config_logging()
+    compile_protobuf()
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/compile_protobuf.py
+++ b/compile_protobuf.py
@@ -4,9 +4,11 @@
 import glob
 import logging
 import os
+import platform
 import shutil
 import sys
 import tempfile
+import time
 import urllib
 import urllib.request
 from subprocess import check_call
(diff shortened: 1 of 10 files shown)


ndodda-amazon (changes requested):
Remove the protobuf installation script and remove [this line](https://github.com/awslabs/sagemaker-debugger/blob/master/config/buildspec_xgboost_1_2_1.yml#L36) from all of the buildspecs. 

Also, talk to @mchoi8739 in order to sanity check whether a RTD build will succeed with these changes. Even though this is in Python, the changes are a high level wrapper over shell commands which may not be allowed in the build environment.

NihalHarish:
> Remove the protobuf installation script and remove [this line](https://github.com/awslabs/sagemaker-debugger/blob/master/config/buildspec_xgboost_1_2_1.yml#L36) from all of the buildspecs.
> 
> Also, talk to @mchoi8739 in order to sanity check whether a RTD build will succeed with these changes. Even though this is in Python, the changes are a high level wrapper over shell commands which may not be allowed in the build environment.

The RTD build seems to be working fine with this change. I will go ahead and update all the buildspec files.

config/buildspec_build_wheel.yml
```diff
@@ -14,7 +14,7 @@ phases:
         - apt-get install sudo -qq -o=Dpkg::Use-Pty=0
         - sudo apt-get update -qq -o=Dpkg::Use-Pty=0
         - sudo apt-get install unzip -qq -o=Dpkg::Use-Pty=0
-        - cd $CODEBUILD_SRC_DIR && chmod +x config/protoc_downloader.sh && ./config/protoc_downloader.sh
+        - cd $CODEBUILD_SRC_DIR
```
ndodda-amazon, line 17: Nothing else in the install stage requires this directory, can we remove this line entirely?

Same for all of the other buildspecs too.

compile_protobuf.py
```diff
@@ -0,0 +1,92 @@
...
+from zipfile import ZipFile
+
+
+def script_name() -> str:
+    """:returns: script name with leading paths removed"""
+    return os.path.split(sys.argv[0])[1]
+
+
+def config_logging():
+    import time
```
ndodda-amazon, line 22: nit: Move this to the imports above.

compile_protobuf.py
```diff
@@ -0,0 +1,92 @@
...
+from subprocess import check_call
+from zipfile import ZipFile
+
+
+def script_name() -> str:
+    """:returns: script name with leading paths removed"""
+    return os.path.split(sys.argv[0])[1]
+
+
+def config_logging():
```
ndodda-amazon, line 21: nit: configure_logging?

compile_protobuf.py
```diff
@@ -0,0 +1,92 @@
...
+    logging.getLogger().setLevel(logging.INFO)
+    logging.basicConfig(format="{}: %(asctime)sZ %(levelname)s %(message)s".format(script_name()))
+    logging.Formatter.converter = time.gmtime
+
+
+def _protoc_bundle():
+    """
+    Returns an archive with the binary protoc distro for the platform
+    """
+    import platform
```
ndodda-amazon, line 33: nit: Move this to the imports above.

compile_protobuf.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def get_protoc():
+    """make sure protoc is available, otherwise download it and return a tuple with the protoc
+    binary and a temporary dir if it needed to be downloaded"""
+    if not shutil.which("protoc"):
+        archive_url = _protoc_bundle()
+        if not archive_url:
+            raise RuntimeError(
+                "protoc not installed. Please install it manually by running sh protoc_downloader.sh"
```
ndodda-amazon, line 55: Can we make this error message more clear? I.e. installation of protoc failed for this OS?

compile_protobuf.py
```diff
@@ -0,0 +1,92 @@
...
+
+def config_logging():
+    import time
+
+    logging.getLogger().setLevel(logging.INFO)
+    logging.basicConfig(format="{}: %(asctime)sZ %(levelname)s %(message)s".format(script_name()))
+    logging.Formatter.converter = time.gmtime
+
+
+def _protoc_bundle():
```
ndodda-amazon, line 29: nit: Rename to `get_protoc_download_url` or something more descriptive.

compile_protobuf.py
```diff
@@ -0,0 +1,92 @@
...
+def get_protoc():
+    """make sure protoc is available, otherwise download it and return a tuple with the protoc
+    binary and a temporary dir if it needed to be downloaded"""
+    if not shutil.which("protoc"):
+        archive_url = _protoc_bundle()
+        if not archive_url:
+            raise RuntimeError(
+                "protoc not installed. Please install it manually by running sh protoc_downloader.sh"
+            )
+        logging.info("Downloading protoc")
```
ndodda-amazon, line 57: Is this logging message necessary since there are already logging messages in `_protoc_bundle`?

compile_protobuf.py
```diff
@@ -0,0 +1,92 @@
...
+        (fname, headers) = urllib.request.urlretrieve(archive_url)
+        tmpdir = tempfile.mkdtemp(prefix="protoc_smdebug")
+        with ZipFile(fname, "r") as zipf:
+            zipf.extractall(tmpdir)
+        protoc_bin = os.path.join(tmpdir, "bin", "protoc")
+        os.chmod(protoc_bin, 0o755)
```
ndodda-amazon, line 63: nit: This makes the the binary executable? Add a comment

compile_protobuf.py
```diff
@@ -0,0 +1,92 @@
...
+        logging.info("Downloading protoc for Linux: %s", archive_url)
+    else:
+        archive_url = None
+    return archive_url
+
+
+def get_protoc():
+    """make sure protoc is available, otherwise download it and return a tuple with the protoc
+    binary and a temporary dir if it needed to be downloaded"""
+    if not shutil.which("protoc"):
```
ndodda-amazon, line 51: Restructure this condition to return early if protoc is already available.

```
if not shutil.which("protoc"):
    return (shutil.which("protoc"), None)

archive_url = ...
...
    
```

This makes the function easier to read.

setup.py
```diff
@@ -36,18 +38,41 @@
...
+        yield
+    finally:
+        os.chdir(curdir)
+
+
+def scan_git_secrets():
+    from subprocess import check_call
+    import os
+    import tempfile
+    import shutil
```
ndodda-amazon, line 57: nit: Can we move these to the imports above?

setup.py
```diff
@@ -36,18 +38,41 @@
...
+
+    if shutil.which("git-secrets"):
+        check_call(["git", "secrets", "--scan"])
+        print("scanned for git secrets")
+
+    else:
+        with tempfile.TemporaryDirectory(prefix="git_secrets_") as tmpdir:
+            check_call(["git", "clone", "https://github.com/awslabs/git-secrets.git", tmpdir])
+            check_call([os.path.join(tmpdir, "git-secrets"), "--scan"])
+        print("scanned for git secrets")
```
ndodda-amazon, line 71: nit: Move this outside of the if else and remove line 65

setup.py
```diff
@@ -36,18 +38,41 @@
...
-    proto_paths = ["smdebug/core/tfevent/proto"]
-    cmd = "set -ex && protoc "
-    for proto_path in proto_paths:
-        proto_files = os.path.join(proto_path, "*.proto")
-        cmd += proto_files + " "
-        print("compiling protobuf files in {}".format(proto_path))
-    cmd += " --python_out=."
-    return os.system(cmd)
+@contextlib.contextmanager
+def remember_cwd():
```
ndodda-amazon, line 42: Where is this used?

setup.py
```diff
@@ -36,18 +38,41 @@
...
+    """
+    curdir = os.getcwd()
+    try:
+        yield
+    finally:
+        os.chdir(curdir)
+
+
+def scan_git_secrets():
+    from subprocess import check_call
```
ndodda-amazon, line 54: What is the behavior if the command executed by `check_call` fails? Does it throw an error?

setup.py
```diff
@@ -36,18 +38,41 @@
...
+    Restore current directory when exiting context
+    """
+    curdir = os.getcwd()
+    try:
+        yield
+    finally:
+        os.chdir(curdir)
+
+
+def scan_git_secrets():
```
ndodda-amazon, line 53: Does this get used? I see a `scan_git_secrets` function redefined below.

compile_protobuf.py
```diff
@@ -0,0 +1,92 @@
...
+    system = platform.system()
+    machine = platform.machine()
+    if system == "Darwin":
+        archive_url = "https://github.com/protocolbuffers/protobuf/releases/download/v3.11.4/protoc-3.11.4-osx-x86_64.zip"
+        logging.info("Downloading protoc for Darwin: %s", archive_url)
+    elif system == "Linux":
+        archive_url = f"https://github.com/protocolbuffers/protobuf/releases/download/v3.11.4/protoc-3.11.4-linux-{machine}.zip"
+        logging.info("Downloading protoc for Linux: %s", archive_url)
+    else:
+        archive_url = None
```
ndodda-amazon, line 44: nit: Perhaps we should just move the Runtime Error here instead of `get_protoc` since there's no reason to propagate the error back up?

ndodda-amazon (changes requested):
Changes look good, left comments around organization as well as nits.

compile_protobuf.py
NihalHarish: removed.

compile_protobuf.py
```diff
@@ -0,0 +1,101 @@
...
+        shutil.rmtree(tmpdir, ignore_errors=True)
+
+
+def main():
+    configure_logging()
+    compile_protobuf()
+    return 0
+
+
+if __name__ == "__main__":
```
ndodda-amazon, line 100: Are this function and `main` necessary? Seems like `compile_protobuf` gets called directly via setup.py and this script doesn't get executed on its own.

Also, two more questions (I'll leave the decision up to you):
- Should this file go under the `config` folder?
- Is a separate file necessary? Can we put these functions in `setup.py` directly?

compile_protobuf.py
```diff
@@ -0,0 +1,101 @@
...
+from subprocess import check_call
+from zipfile import ZipFile
+
+
+def script_name() -> str:
+    """:returns: script name with leading paths removed"""
+    return os.path.split(sys.argv[0])[1]
+
+
+def configure_logging():
```
ndodda-amazon, line 23: nit: `setup.py` uses print statements and we use logging here. To be more consistent, can we make everything logging statements (preferable) or print statements?

ndodda-amazon (approved):
LGTM, thanks for fixing this! Left a few minor comments, feel free to merge after addressing them.

PR #525, opened 2022-01-27
Disable gradient tape profiling tests

adimux:
### Description of changes:
Tests related to profiling using gradient tape are failing
and blocking the TF2.7 release.
Since this feature has never been released, we can safely
disable the tests, until we make a decision about supporting this
feature.

#### Style and formatting:


tests/profiler/tensorflow2/test_native_tf2_profiler.py
```diff
@@ -146,7 +146,10 @@ def _verify_timeline_files(out_dir):
             "process_sort_index",
         }
 
-
+# Skipping because the tests are failing.
+# Support for profiling using gradient tape has never been released publicly
+# and since we're planning on deprecating profiler v1, we can just disable the tests
+@pytest.mark.skip(reason="unblock TF2.7 release")
```
ndodda-amazon, line 152: The tests only fail for TF 2.7, right? Then let's use `skipif` to skip the test for all TF versions 2.7 and above.

ndodda-amazon (changes requested):
If profiler_v1 is deprecated, I would actually just recommend removing [this line](https://github.com/awslabs/sagemaker-debugger/blob/master/config/tests.sh#L37). So you don't need to manually skip each test in this file (and you can ignore the other comment I made).

adimux:
xgboost tests are failing in the nightly build of `sagemaker-debugger`, so this PR is ok to merge

ndodda-amazon (approved):
LGTM, great work!

josephevans (approved):
Thanks, LGTM.

tests/profiler/tensorflow2/test_native_tf2_profiler.py
hsl89: Hi @ndodda-amazon, do tests under `tests/profiler/tensorflow2/` depends on the code changes applied to the tf code base?

Roshrini:
LGTM. Thanks

Roshrini (approved):

PR #526, opened 2022-01-31
Compatibility with TF 2.7 

adimux:
...by fixing broken tests in test_keras_mirrored.py

Motive: Tensorflow 2.7.0 release

`smdebug.tensorflow.get_hook()` was returning `None`, which made many
tests fail, but this was actually a false alarm.
We just have to mark the version of tensorflow we are trying to release (`2.7.0`)
as a supported framework version, so `get_hook()` doesn't return
`None'...

I know what you think. It's not pythonic to silently fail and we should
instead throw an exception, but this code preceded me and I don't want
to introduce a breaking change (an exception being raised) and I don't
want to introduce warnings either in case `get_hook()' gets called
in unexpected ways multiple times during the lifecycle of a training job.

Instead, I just created a tests that alarms us of the situation
when `get_hook()` might return `None` when it's actually just because we forgot to update a constant.

#### Style and formatting:

### DIFF
--- a/smdebug/core/utils.py
+++ b/smdebug/core/utils.py
@@ -653,5 +653,8 @@ def is_framework_version_supported(framework_type):
         from smdebug.tensorflow.utils import is_current_version_supported
 
         return is_current_version_supported()
-    else:  # If framework is mxnet and XGBoost we will currently return True for all versions
+    # If framework is mxnet and XGBoost we will currently return True for all versions
+    elif framework_type in [FRAMEWORK.XGBOOST, FRAMEWORK.MXNET]:
         return True
+
+    raise Exception(f"No such framework type {framework_type}")
--- a/tests/tensorflow/keras/test_keras.py
+++ b/tests/tensorflow/keras/test_keras.py
@@ -12,6 +12,7 @@ from smdebug.core.access_layer import has_training_ended
 from smdebug.core.collection import CollectionKeys
 from smdebug.core.modes import ModeKeys
 from smdebug.core.reduction_config import ALLOWED_NORMS, ALLOWED_REDUCTIONS
+from smdebug.core.utils import FRAMEWORK, is_framework_version_supported
 from smdebug.exceptions import TensorUnavailableForStep
 from smdebug.tensorflow import ReductionConfig, SaveConfig
 from smdebug.tensorflow.keras import KerasHook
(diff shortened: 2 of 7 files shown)


ndodda-amazon (changes requested):
What framework version of TF is being used in the smdebug_tf2_build? Can you confirm that it's using TF 2.7?

adimux:
Our pipeline runs with an outdated version of Tensorflow (`2.4`). I will fix the pipeline later, but right now we need to unblock the release of `2.7`.
Fortunately, the DLC pipelines do run the sagemaker-debugger tests and I also ran them locally and they're green.

hsl89:
LGTM

Roshrini (commented):
Did you test it in TF2.7 DLC built through DLC pipeline?

Roshrini (approved):

abhinavs95:
LGTM

adimux:
created other PR to be able to merge

smdebug/core/utils.py
```diff
@@ -653,5 +653,8 @@ def is_framework_version_supported(framework_type):
         from smdebug.tensorflow.utils import is_current_version_supported
 
         return is_current_version_supported()
-    else:  # If framework is mxnet and XGBoost we will currently return True for all versions
+    # If framework is mxnet and XGBoost we will currently return True for all versions
+    elif framework_type in [FRAMEWORK.XGBOOST, FRAMEWORK.MXNET]:
         return True
+
+    raise Exception(f"No such framework type {framework_type}")
```
olegtalalov, line 660: This should be ValueError exception.

tests/tensorflow/keras/test_keras.py
```diff
@@ -135,6 +136,25 @@ def train_model(
...
+    This test is designed to save you 2 days of debugging time in case you're new
+    to the codebase.
+
+    One of the side-effects of not updating LATEST_SUPPORTED_TF_VERSION with the
+    version of Tensorflow you're trying to release for is that `smdebug.tensorflow.get_hook()` is going to return `None`
+    and many tests will fail.
+
+    This is just to make you aware of the problem.
+    """
+    if not is_framework_version_supported(FRAMEWORK.PYTORCH):
```
olegtalalov, line 150: Why PYTORCH?

PR #544, opened 2022-03-04
Move zero code change tests

mariumof:
#### Style and formatting:


adimux:
get_hook() is always used, not only for zero-code-change.
Even when the hooks is manually created, it's registered in a global variable and at the end of the training phase, I'm pretty sure `get_hook().cleanup()` gets called.

johnbensnyder (approved):

PR #572, opened 2022-03-19
Add the option to remove stable_release env var, none of the other bu…

mariumof:
…ild specs use it

#### Style and formatting:

### DIFF
--- a/config/buildspec_vanilla_framework_tests.yml
+++ b/config/buildspec_vanilla_framework_tests.yml
@@ -22,13 +22,13 @@ phases:
         - sudo apt-get install unzip -qq -o=Dpkg::Use-Pty=0
         - pip install --upgrade pip==20.3.3
         - pip install -q -U pytest pytest-cov wheel pyYaml pytest-html keras==2.3.1 mxnet==1.6.0 torch==1.8.0 xgboost==1.3.3 pre-commit tensorflow_datasets==4.0.1 torchvision ipython_genutils==0.2.0
-        - cd $CODEBUILD_SRC_DIR && chmod +x config/install_smdebug.sh && chmod +x config/check_smdebug_install.sh && ./config/install_smdebug.sh;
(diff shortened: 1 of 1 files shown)


MZSHAN (approved):

config/buildspec_vanilla_framework_tests.yml
```diff
@@ -22,13 +22,13 @@ phases:
         - sudo apt-get install unzip -qq -o=Dpkg::Use-Pty=0
         - pip install --upgrade pip==20.3.3
         - pip install -q -U pytest pytest-cov wheel pyYaml pytest-html keras==2.3.1 mxnet==1.6.0 torch==1.8.0 xgboost==1.3.3 pre-commit tensorflow_datasets==4.0.1 torchvision ipython_genutils==0.2.0
-        - cd $CODEBUILD_SRC_DIR && chmod +x config/install_smdebug.sh && chmod +x config/check_smdebug_install.sh && ./config/install_smdebug.sh;
+        - if [ "$stable_release" ]; then echo "Installing from S3" && cd $CODEBUILD_SRC_DIR && chmod +x config/install_smdebug.sh && chmod +x config/check_smdebug_install.sh && ./config/install_smdebug.sh; else echo "Installing NOT from S3" && cd $CODEBUILD_SRC_DIR  && python setup.py bdist_wheel --universal && cd $CODEBUILD_SRC_DIR_RULES && python setup.py bdist_wheel --universal && cd $CODEBUILD_SRC_DIR  && pip install --force-reinstall dist/*.whl && cd ..; fi
```
galaind, line 25: how do you specify stable_release?

Also, when you specify 'Installing from S3' would it be possible to specify the S3 path instead?

mariumof: environment variable. It means it automatically picks up from the S3 dumps of the other builds. If the other builds did not finish, then no S3 to pick up from (they are by date). The way I see it, it is a relic, but I kept it, just in case
... Re would it be possible? Yes, but the antiquated install.sh is convoluted enough as it is. Should we invest in this?

johnbensnyder (approved):

PR #573, opened 2022-03-24
Debug

mariumof:
#### Style and formatting:


config/buildspec_tensorflow_2_5.yml
```diff
@@ -0,0 +1,64 @@
+# Build Spec for AWS CodeBuild CI TF 2.3.1 CPU and GPU Containers
```
Roshrini, line 1: buildspec is for TF 2.5, these comments point to TF 2.3.1
Are we still using 2.3 as target containers? Can you verify?

PR #592, opened 2022-05-10
add more descriptive message when json reading error happens

zaoliu-aws:
### Description of changes:
 Print more informative message when json reading error occurs. Right now, errors like https://t.corp.amazon.com/V556254533 is reoccurring. By adding these extra information, I could find out if these error happens (1) with the last json file (2) the first json file (3) any json file. I hope it turns out to be case (1).

### DIFF
--- a/smdebug/profiler/system_profiler_file_parser.py
+++ b/smdebug/profiler/system_profiler_file_parser.py
@@ -1,6 +1,7 @@
 # First Party
 # Standard Library
 import json
+import os
 
 from smdebug.core.logger import get_logger
 from smdebug.profiler.utils import TimeUnits, convert_utc_timestamp_to_nanoseconds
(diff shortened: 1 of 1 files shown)


zaoliu-aws:
@johnbensnyder, @jleeleee , please take a look.

johnbensnyder (approved):

zaoliu-aws:
@jleeleee , could you please also take a look?

smdebug/profiler/system_profiler_file_parser.py
```diff
@@ -51,16 +52,34 @@ def read_events_from_file(self, eventfile):
...
+            errStr = 'line No. ' + str(lineNum) + ' out of total ' + str(lineNumTotal) + ' lines, '
+            cur_dir = os.path.dirname(eventfile)
+            cur_dir_base = os.path.basename(cur_dir)
+            eventfile_base = os.path.basename(eventfile)
+            parent_dir = os.path.dirname(cur_dir)
+            time_stamp_files = [int(x.split('.')[0]) for x in os.listdir(cur_dir)]
+            time_stamp_dirs = [int(x) for x in os.listdir(parent_dir)]
+            file_pos = time_stamp_files.index(int(eventfile_base.split('.')[0]))
+            dir_pos = time_stamp_dirs.index(int(cur_dir_base))
+            errStr += 'file No.' + str(file_pos+1) + ' in file No.1 to No.' + str(len(time_stamp_files)) + ', folder No.' + str(dir_pos+1) + " in folder No.1 to No." + str(len(time_stamp_dirs)) + "\n"
```
jleeleee, line 77: This is code that's publicly exposed to all customers. We should make the error message a lot more descriptive, and also, instead of doing `len(time_stamp_dirs)`,  just set the original variable as the length

smdebug/profiler/system_profiler_file_parser.py
```diff
@@ -51,16 +52,34 @@ def read_events_from_file(self, eventfile):
...
             self.logger.error(
-                f"Can't open profiler system metric file {eventfile}: Exception {str(e)}"
+                f"{errStr} Can't open profiler system metric file {eventfile}: Exception {str(e)}"
```
jleeleee, line 79: Don't print this at the start of the message, print it somewhere in the middle or the end, so that customers can quickly categorize the error

smdebug/profiler/system_profiler_file_parser.py
jleeleee: actually nvm, the len(time_stamp_dirs) is fine, but the string isn't very descriptive.
Also, nit but this line of code is too long. Can you break it up into separate lines?

zaoliu-aws: pushed in changes to make the code more descriptive. Also, breaking up the long line.

ZiweiG (approved):

PR #597, opened 2022-05-23
make the processing function more robust to handle corrupted json lin…

zaoliu-aws:
A small change to make the processing function more robust.


zaoliu-aws:
@johnbensnyder and @jleeleee , please take a look to see if it makes sense. It is a very small change.

johnbensnyder (approved):

zaoliu-aws:
added more information based on @jleeleee 's input.

jleeleee (approved):
nit: probably could be debug level, but shouldn't matter too much

zaoliu-aws:
@jleeleee , changed to debug level.

PR #604, opened 2022-06-28
add a check to determine if horovod.torch import succeeds

zaoliu-aws:
import horovod.torch does not raise exception even if it is not fully successful. Add init() function to catch it.

### DIFF
--- a/smdebug/core/utils.py
+++ b/smdebug/core/utils.py
@@ -64,14 +64,19 @@ try:
 except (ImportError, ModuleNotFoundError):
     _torch_dist_imported = None
 
-
 try:
     import horovod.torch as hvd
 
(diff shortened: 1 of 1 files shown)


smdebug/core/utils.py
```diff
@@ -64,24 +64,29 @@ class FRAMEWORK(Enum):
...
 
-
+logger = get_logger()
 try:
     import horovod.torch as hvd
 
     # This redundant import is necessary because horovod does not raise an ImportError if the library is not present
     import torch  # noqa
+    #make sure the library is correctly imported
+    hvd.init()
```
johnbensnyder, line 74: Do we know if this could interfer with other distribution strategies? I'm worried running init on horovod might cause a problem if the user is running SMDDP or DDP.

PR #611, opened 2022-08-15
Core logging

jleeleee:
### Description of changes:
Added new custom exception type that will be used specifically for all SM Debugger related exceptions

#### Style and formatting:

N/A

### DIFF
--- a/smdebug/exceptions.py
+++ b/smdebug/exceptions.py
@@ -1,8 +1,10 @@
 # First Party
 from smdebug.core.modes import ModeKeys as modes
 
+
 class SMDebugError(Exception):
     pass
+
 
 class SMDebugCustomerError(SMDebugError):
     pass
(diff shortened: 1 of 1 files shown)


josephevans (approved):

smdebug/exceptions.py
```diff
@@ -2,6 +2,14 @@
 from smdebug.core.modes import ModeKeys as modes
 
 
+class SMDebugError(Exception):
+    pass
+
+
+class SMDebugCustomerError(SMDebugError):
```
yl-to, line 9: when will we have a customer error?

jleeleee: this will be reserved for things like incorrect hook configuration (e.g. missing hook initialization on training running on EC2 rather than sagemaker, since sagemaker automatically creates a JSON config but EC2 won't).
For now, we'll classify all errors as SMDebugError and fill this in as necessary.

yl-to: LGTM

yl-to (approved):
Approving

PR #613, opened 2022-08-22
Refactored MXNet/PyTorch/TF Exceptions, and End-of-training log directory check

jleeleee:
### Description of changes:
Refactored all exception types under the `smdebug/mxnet`, `smdebug/pytorch`, and `smdebug/tensorflow` directories to return the `SMDebugError` exception type. Also changed the logic of the end of training log directory existence check in `smdebug/core/hook_utils.py` for the writer to return a boolean instead of raising an exception that gets caught later on, to make it clear that exceptions are generally meant to flow up to the customer, not to be used as a check.

#### Style and formatting:

### DIFF
--- a/smdebug/core/access_layer/utils.py
+++ b/smdebug/core/access_layer/utils.py
@@ -34,12 +34,9 @@ def training_has_ended(trial_prefix):
             f"The end of training job file will not be written for jobs running under SageMaker."
         )
         return
-    try:
-        check_dir_exists(trial_prefix)
-        # if path does not exist, then we don't need to write a file
-    except RuntimeError:
-        # dir exists
-        pass
+
+    check_dir_exists(trial_prefix)
+
     file_path = os.path.join(trial_prefix, END_OF_JOB_FILENAME)
     s3, bucket_name, key_name = is_s3(file_path)
     if s3:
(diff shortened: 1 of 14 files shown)


yl-to (approved):
LGTM

smdebug/core/access_layer/utils.py
```diff
@@ -34,12 +34,9 @@ def training_has_ended(trial_prefix):
...
         )
         return
-    try:
-        check_dir_exists(trial_prefix)
-        # if path does not exist, then we don't need to write a file
-    except RuntimeError:
-        # dir exists
-        pass
+
+    check_dir_exists(trial_prefix)
```
dkey-amazon, line 38: What is the purpose of this line, since the return value appears to be ignored and no error is raised?  As far as I can tell, this `check_dir_exists(trial_prefix)` check doesn't do anything, correct?  Should it maybe be removed, or made part of an `if` clause?

jleeleee: I was personally debating that as well. `check_dir_exists` performs a call to S3, and would raise an exception if the exception wasn't just `NoSuchBucket` (e.g. invalid credentials, etc.). This would in turn skip the below logic for the writer, which could have been part of the desired impact from  the previous implementation (where it basically said `if dir exists, do nothing, if dir doesn't exist, do nothing, otherwise, catch some other exceptions`).

So ultimately, it's an early termination of the function in case there's some S3 config issues, which isn't completely necessary, but helpful

dkey-amazon: Ok, that makes sense.  I'd maybe just add a comment explaining that then, for the sake of readability.

jleeleee: added a short comment

dkey-amazon (approved):
LGTM, thanks for the clarifying comment @jleeleee

PR #619, opened 2022-09-02
fix SyntaxWarning caused by using is not

zaoliu-aws:
### Description of changes:
Comparing strings using 'is not' causes syntax warning. 
SyntaxWarning: "is not" with a literal. Did you mean "!="?


jleeleee (approved):

dkey-amazon (approved):
LGTM

PR #621, opened 2022-09-09
Refactored core directory exceptions

jleeleee:
### Description of changes:
Refactored core directory exceptions to inherit from `SMDebugError` exceptions for telemetry in debugger-enabled SageMaker Training jobs. Also refactored asserts to explicit exception throws.

#### Style and formatting:


smdebug/exceptions.py
```diff
@@ -10,15 +10,23 @@ class SMDebugCustomerError(SMDebugError):
...
-class InvalidCollectionConfiguration(Exception):
+class SMDebugNotImplementedError(SMDebugError, NotImplementedError):
+    pass
+
+
+class SMDebugValueError(SMDebugError, ValueError):
+    pass
+
+
+class InvalidCollectionConfiguration(SMDebugError):
```
yl-to, line 29: Are we going to add `SMDebug` before `InvalidCollectionConfiguration` here?

jleeleee: seems unnecessary since the telemetry is going to catch the exceptions from the class SMDebugError, not from the name of the exception

yl-to (approved):

josephevans (approved):

PR #632, opened 2022-12-12
fix: embedding layer grad collecting

yl-to:
#### Style and formatting:


smdebug/tensorflow/keras.py
```diff
@@ -1239,12 +1239,19 @@ def run(*args, **kwargs):
                         and hasattr(vars[0], "numpy")
                     )
                 )
-                or (not ((isinstance(grads[0], tf.Tensor)) and hasattr(grads[0], "numpy")))
             ):
                 return grads
+            # get embedding layers indexes
+            embedding_grad_indexes = set()
+            for idx, grad in enumerate(grads):
```
johnbensnyder, line 1246: Can we expect that all gradients in the list are of the same type? In which case, can we stick with the previous logic of checking the first element? If they're not all the same type, this might still miss some cases if it returns grads before checking all elements.

smdebug/tensorflow/keras.py
```diff
@@ -1239,12 +1239,19 @@ def run(*args, **kwargs):
...
+                if not (isinstance(grad, tf.Tensor) and hasattr(grad, "numpy")):
+                    if isinstance(grad, tf.IndexedSlices):
+                        embedding_grad_indexes.add(idx)
+                    else:
+                        return grads
```
olegtalalov, line 1251: This logic and using of `set` is hard to read and understand. It can be simplified: 
```
# if we have unknown object type then do nothing
if any(map(lambda grad: not (is_valid_tensor(grad) ||  is_IndexedSlices(grad), grads)):
    return grads
```

smdebug/tensorflow/keras.py
```diff
@@ -1253,6 +1260,8 @@ def run(*args, **kwargs):
                     # model.variables includes trainable and non-trainable
                     # variables.
                     if g is not None:
+                        if idx in embedding_grad_indexes:
```
olegtalalov, line 1263:
```
if is_IndexedSlices(g):
    g = tf.convert_to_tensor(g)
```

olegtalalov (changes requested):
Overall feedback, from non debuger member: this change mixes together some logic (return if a list of grads contained unknown object) and some processing (collection IndexedSlices). Usually there is no reason to do this together. Maybe I missed something, but implementation using sets is not required - it is just an additional complexity.

smdebug/tensorflow/keras.py
```diff
@@ -1239,12 +1241,18 @@ def run(*args, **kwargs):
...
+            if any(
+                    map(
+                        lambda grad: not (is_valid_tf_tensor(grad) or is_IndexedSlices(grad)), grads
+                    )
             ):
                 return grads
 
             if self._get_collections_to_save_for_step():
-                for (g, v) in zip(grads, vars):
+                for idx, (g, v) in enumerate(zip(grads, vars)):
```
olegtalalov, line 1255: idx is not used anymore

smdebug/tensorflow/utils.py
```diff
@@ -412,6 +412,17 @@ def get_keras_mode(mode):
         return KerasModeKeys.PREDICT
 
 
+def is_valid_tf_tensor(grad):
+    if isinstance(grad, tf.Tensor) and hasattr(grad, "numpy"):
+        return True
+    else:
+        return False
```
olegtalalov, line 419:
```
    return isinstance(grad, tf.Tensor) and hasattr(grad, "numpy")
```

olegtalalov (approved):
I left few comments, but overall looks good.

johnbensnyder (approved):

PR #636, opened 2022-12-20
skip trace.json file reader for tf 2.11 and above since it's no longe…

ztlevi:
### Description of changes:
TF 2.11 no longer have trace.json.gz file available. They merged all data to `xxx-.xplane.pb` file. This PR disables TF framework profiler support for TF 2.11 and above to unblock the AWS TF release.

#### Style and formatting:

### DIFF
--- a/tests/profiler/tensorflow2/test_tf2_profiler.py
+++ b/tests/profiler/tensorflow2/test_tf2_profiler.py
@@ -5,6 +5,9 @@ import os
 import pytest
 from tests.profiler.tensorflow2.utils import verify_detailed_profiling
 from tests.tensorflow2.test_keras import helper_keras_fit
+import tensorflow as tf
+from packaging import version
+
 
 # First Party
 from smdebug.profiler.profiler_config_parser import ProfilerConfigParser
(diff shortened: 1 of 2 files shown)


tests/profiler/core/test_utils.py
```diff
@@ -138,7 +142,8 @@ def test_timeline_merge_by_step(out_dir, trace_location):
 
     assert total_events == num_trace_events
 
-
+@pytest.mark.skipif(version.parse(tf.__version__) > version.parse("2.10"),
+                    reason="tf 2.11 merged trace.json file to xplane.pb file.")
```
Roshrini, line 146: why do you need to skip all tests related to merging timelines? only one file is not available, rests still should work

MZSHAN (approved):

smdebug/profiler/algorithm_metrics_reader.py
```diff
@@ -4,6 +4,10 @@
 import json
 import os
 
+# Third Party
+import tensorflow as tf
```
yl-to, line 8: what if this test is running in a pytorch container?

ztlevi: you're right. reverting the change from this file

yl-to:
If profiler team's decision is to disable the failing tests, I would suggest `@pytest.mark.skipif` before the pytest func def

ztlevi:
Pushed an update for your comments @yl-to

yl-to:
@ztlevi So profiler team's decision is to skip these tests for tensorflow 2.11, please confirm after you tested it.
I will approve the PR once get confirmation.

ztlevi:
Yes, I confirm tests passes.

yl-to (approved):

PR #639, opened 2022-12-27
fix: numpy deprecated aliases

ShiboXing:
#### Style and formatting:

### DIFF
--- a/smdebug/core/hook.py
+++ b/smdebug/core/hook.py
@@ -734,7 +734,7 @@ class BaseHook:
                     continue
                 elif s_col.save_histogram is True:
                     np_value = self._make_numpy_array(tensor_value)
-                    if self.dry_run or np_value.dtype == np.bool_ or np_value.nbytes == 0:
+                    if self.dry_run or np_value.dtype == bool or np_value.nbytes == 0:
                         return
 
                     hist_name = f"{s_col.name}/{tensor_name}"
--- a/smdebug/core/hook.py
+++ b/smdebug/core/hook.py
@@ -734,7 +734,7 @@ class BaseHook:
                     continue
                 elif s_col.save_histogram is True:
                     np_value = self._make_numpy_array(tensor_value)
-                    if self.dry_run or np_value.dtype == np.bool or np_value.nbytes == 0:
+                    if self.dry_run or np_value.dtype == np.bool_ or np_value.nbytes == 0:
                         return
 
                     hist_name = f"{s_col.name}/{tensor_name}"
--- a/smdebug/core/tfevent/summary.py
+++ b/smdebug/core/tfevent/summary.py
@@ -92,7 +92,7 @@ def _make_histogram(values, bins, max_bins):
         limits = new_limits
 
     # Find the first and the last bin defining the support of the histogram:
-    cum_counts = np.cumsum(np.greater(counts, 0, dtype=np.int32))
+    cum_counts = np.cumsum(np.greater(counts, 0))
     start, end = np.searchsorted(cum_counts, [0, cum_counts[-1] - 1], side="right")
     start = int(start)
     end = int(end) + 1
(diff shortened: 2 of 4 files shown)


jleeleee (approved):

smdebug/core/hook.py
```diff
@@ -734,7 +734,7 @@ def _write_histogram_summary(self, tensor_name, tensor_value, save_collections):
                     continue
                 elif s_col.save_histogram is True:
                     np_value = self._make_numpy_array(tensor_value)
-                    if self.dry_run or np_value.dtype == np.bool or np_value.nbytes == 0:
+                    if self.dry_run or np_value.dtype == np.bool_ or np_value.nbytes == 0:
```
jleeleee, line 737: Is there a reason we can't use the built in types instead? That seems to be recommended by numpy https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

jleeleee: actually, I guess we don't have good enough knowledge of test coverage to know whether we can use built-in types or not

ShiboXing:
I've tested with the newest numpy that the following checks are all True.
`
    import numpy as np
    test = np.array([True, False])
    
    test.dtype == np.bool_
    test.dtype == bool

    test = np.array([1.0, 2.0])
    test.dtype == float
    test.dtype == np.float_
    np.float_ == np.float64
`

However, I think using np.float_ and np.bool_ instead of native python data types could be safer due to potential future changes by numpy. What do you think?

yl-to:
> I've tested with the newest numpy that the following checks are all True.
> 
> ```
> import numpy as np
> test = np.array([True, False])
> 
> test.dtype == np.bool_
> test.dtype == bool
> 
> test = np.array([1.0, 2.0])
> test.dtype == float
> test.dtype == np.float_
> np.float_ == np.float64
> ```
> 
> However, I think using np.float_ and np.bool_ instead of native python data types could be safer due to potential future changes by numpy. What do you think?

I tested: `np.bool_ == bool`, it returned `False`, are they different?

ShiboXing:
Yeah they are the same when they are compared to the dtype of a np.array separately. But their classes are impelmented differently. 

I just realized that dtypes are returned from this mapping: https://github.com/awslabs/sagemaker-debugger/blob/ff3b53607c36008164372b4fa785545f2e3ec2ef/smdebug/core/tfevent/event_file_reader.py#L42
Here the bool is python native. So I think it's better to stick with Python native for if check as well?

yl-to:
> 

The deprecation notice from `numpy` is like this:
```
DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. 
To silence this warning, use `float` by itself. 
Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
```
So I think we would like to follow `numpy`'s suggestion to make safe changes.

smdebug/core/tfevent/summary.py
```diff
@@ -92,7 +92,7 @@ def _make_histogram(values, bins, max_bins):
         limits = new_limits
 
     # Find the first and the last bin defining the support of the histogram:
-    cum_counts = np.cumsum(np.greater(counts, 0, dtype=np.int32))
+    cum_counts = np.cumsum(np.greater(counts, 0))
```
johnbensnyder, line 95: Can we keep an int dtype here, or is the output of np.cumsum always an int? Want to make sure we have a consistent type.

johnbensnyder (approved):

smdebug/core/tfevent/summary.py
ShiboXing: I've tested that `np.cumsum(np.greater(counts, 0))` outputs int64 regardless if inputs are of type bool or int32.

PR #642, opened 2023-01-16
Rulelet backend

MZSHAN:
#### Style and formatting:


mariumof (commented):
I think this PR should be closed; I will be opening another PR which moves network and memory into the existing collection/storage mechanism. That is needed for many reasons, can explain.

PR #651, opened 2023-03-06
fix: smp will not be imported if not specified by user

yl-to:
### Description of changes:
This PR is created as part of fixing a COE issue in PyTorch 

#### Style and formatting:


smdebug/core/utils.py
```diff
@@ -49,18 +49,21 @@ class FRAMEWORK(Enum):
...
-except (ImportError, ModuleNotFoundError):
+if check_smmodelparallel_training():
     try:
-        import smdistributed.modelparallel.torch as smp
+        import smdistributed.modelparallel.tensorflow as smp
 
         _smp_imported = smp
     except (ImportError, ModuleNotFoundError):
-        _smp_imported = None
+        try:
```
ydaiming, line 61: Instead of chaining the two try-except block, can we decouple them?

Currently, if `smp.tensorflow` failed to import, the code tries to import `smp.torch`.
However, if `smp.torch` fails to import with an error other than `(ImportError, ModuleNotFoundError)`, the stack trace will be a little confusing, which happened in this incident of COE (`PermissionError`).

In particular, two stack traces showed up in this COE incident:
1. try import `smp.tf` failed
2. when trying to solve the above, PermissionError shows up
    1. no except block caught this type of error
    2. no pytorch keyword mentioned on the error itself.

You can see this led to wrong investigation in a PyTorch scenario.

smdebug/core/utils.py
```diff
@@ -49,18 +50,18 @@ class FRAMEWORK(Enum):
 _smddp_tf_imported = None
 _smddp_pt_imported = None
 _is_using_smmodelparallel = None
+_smp_imported = None
```
ydaiming, line 53: where is this `_smp_imported` used in the codebase? I can't seem to find any other places that make use of the result.

I only see one place in `smdebug/core/config_validator.py` that uses the `check_smmodelparallel_training()` function.

Maybe, we should remove this part entirely?

yl-to: https://github.com/yl-to/sagemaker-debugger/blob/56fabe531692403e77ce9b5879d55211adec238e/smdebug/core/utils.py#L449
Please check the above line, here is the calling chain `_smp_imported`->`core/utils.py:get_distributed_worker` -> `core/utils.py:get_node_id`, and `get_node_id` is called in several places in core folder.

ydaiming (approved):
LGTM

johnbensnyder (approved):

ydaiming (approved):

smdebug/core/utils.py
```diff
@@ -35,6 +35,7 @@
     SMDebugRuntimeError,
     SMDebugTypeError,
     SMDebugValueError,
+    SMDebugError
```
josephevans, line 38: Is this being used anywhere?

yl-to: yeah, at line 64 in this file

